PR #257, opened 2022-03-13
Add Al Kharid, Wilderness item spawns and mubariz

jarryd229:
added Al Kharid, Wilderness item spawns and added mubariz spawn and dialogue


GregHib (approved):
I've made a few formatting changes as it was easier than submitting a lot of change requests.

1. Method names show start with lowercase character and longer names doesn't necessarily mean more descriptive.
I.e `advice()` > `DoYouHaveAnyAdviceForMe()`

2. Spaces not tabs for indenting
You can set tab to intent with 4 spaces in settings of all ides and editors :)

3. I removed the original runelite id comments as they can still be figured out.
e.g
```kotlin
player("uncertain", "What kind of options are there?")//575 expression anim id from runelite
```
has 1 line or `expression_uncertain1` in animations.yml and `expression_uncertain_old1` is 575.
I can understand wanting to keep the original data though

5. Formatting for raw strings I think it looks better when line are indented by 1 and triple quotes on separate lines

```kotlin
npc("talking", """
    The rules are fairly self-evident with lots of different
    combinations for you to try out!
""")
```
instead of
```kotlin
npc("talking", """The rules are fairly self-evident with lots of different
combinations for you to try out!""")
```

Although that's just my opinion it's fine if it's slightly different.

Aside from those small things thank you and good work :+1:

PR #294, opened 2022-10-14
organized objects.yml and started quests and canoes

jarryd229:
organized objects.yml in id order and added examines

### DIFF
--- a/data/definitions/npcs.yml
+++ b/data/definitions/npcs.yml
@@ -1,56 +1,17 @@
 hans:
   id: 0
-  race: "human"
-  examine: "Servant of the Duke of Lumbridge."
-rat:
-  id: 47
-  att: 1
-  str: 1
-  def: 1
-  mage: 1
-  range: 1
-  stab_def: -42
-  att_bonus: -47
-  str_bonus: -53
-  slash_def: -42
-  crush_def: -42
-  magic_def: -42
-  range_def: -42
-  attack_speed: 4
-  wander_radius: 8
-  attack_radius: 8
-  hitpoints: 20
-  slayer_xp: 2.0
-  max_hit_melee: 1
-  style: "crush"
-  examine: "A popular dwarven delicacy."
-  race: "rat"
-  height: 5
-cow:
-  id: 12362
-  def: 1
-  att: 1
-  str: 1
-  mage: 1
-  range: 1
-  stab_def: -21
-  att_bonus: -15
-  str_bonus: -15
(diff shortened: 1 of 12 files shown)


data/definitions/npcs.yml
```diff
@@ -72,122 +33,60 @@ chicken:
...
-  crush_def: -15
-  magic_def: -15
-  range_def: -15
+  stab_def: -42
+  att_bonus: -47
+  str_bonus: -53
+  slash_def: -42
+  crush_def: -42
+  magic_def: -42
+  range_def: -42
```
GregHib, line 62: Any reason why you update this one and not the others?

data/definitions/npcs.yml
```diff
@@ -72,122 +33,60 @@ chicken:
...
-  crush_def: -26
-  magic_def: -26
-  range_def: -26
+  stab_def: -10
+  att_bonus: -10
+  str_bonus: -10
+  slash_def: -10
+  crush_def: -10
+  magic_def: -10
+  range_def: -10
```
GregHib: Sheep aren't attackable so don't need combat stats 🐑

GregHib:
I'm a bit torn on the objects.yml

## Examines
Pro: it's nice to have them all
Con: The objects.yml file is now massive and that'll slow loading times for future modifications, but it'll inevitably get large the more content is added anyway.

## Sorted by id
Pro: easy to find if you know the id of something, nice and orderly/organised
Con: you lose the context information like pairs of doors 

```
prison_door_closed: 79
prison_door_opened: 43046
```

data/definitions/npcs.yml
jarryd229: looking at the original it was already there

data/definitions/npcs.yml
jarryd229: when using the YamlSorter it's done that it switched some things around that one looks to be from duck_grey: id: 6113 or rat: id: 47

data/definitions/npcs.yml
jarryd229: i've gone through npcs.yml should be fixed

data/definitions/npcs.yml
jarryd229: i've gone through npcs.yml should be fixed and made sure bonus are good

GregHib (approved):
Did a bit of formatting myself and sorted the objects by id + pairs

PR #349, opened 2023-06-13
all of free world stairs and ladders

jarryd229:
should be all of free world stairs and ladders added


data/map/stairs.yml
```diff
@@ -1,1125 +1,1295 @@
...
+    y: 4835
+######## 6989 ########
+- id: 24427
+  option: Walk-up
+  tile:
+    x: 1758
+    y: 4959
+  to:
+    x: 3258
+    y: 3452
```
GregHib:
```suggestion
  delta:
    x: 1500
    y: -1505
```

GregHib (approved):

PR #359, opened 2023-07-12
Add the knights sword quest

jarryd229:
added the knights sword quest


GregHib:
Great, thanks as always!😄 

I made a few changes:
* Fixed thurgo's chat animations
* Renamed variable stages
* Added Sir Vyvin and the cupboard line of sight check

And tidied up choice/option usages, where the first message from the player is the same as the option, you can just add the expression to the option and it'll do it automatically:

```kotlin
option("Well, do you know the VAGUE AREA you lost it in?") {
    player<Unsure>("Well, do you know the VAGUE AREA you lost it in?")
    // ...
}
```

```kotlin
option<Unsure>("Well, do you know the VAGUE AREA you lost it in?") {
    // ...
}
```

PR #436, opened 2024-01-29
Gunnars ground quest

jarryd229:
some animations need naming
itemsbox with 2 items shows but won't continue
cutscene still needs some work
missing Gudrun Hugging Dororan anim
antique_lamp_gunnars_ground needs adding
animation 6709 only workd once in the cutscene


data/definitions/animations.yml
```diff
@@ -1,4 +1,7 @@
 # Dialogue expressions
+
+expression_unknown_expression: 9807
```
GregHib: Is this the correct unknown id? cause it's the same id as `Talk`

jarryd229: no that was just so i can know its wrong and need to come back and fix it i didn't want to put up a wrong expression

GregHib: Have you got the actual id? I'll see if I can find an equivilant

jarryd229: no i don't sorry

game/src/main/kotlin/world/gregs/voidps/world/map/barbarian_village/Haakon.kts
```diff
@@ -0,0 +1,54 @@
...
+    menu()
+}
+
+suspend fun CharacterContext.menu() {
+    npc<Furious>("I am Haakon, champion of this village. Do you seek to challenge me?")
+    choice {
+        option<Talking>("I challenge you!") {
+            attack()
+        }
+        val validStages = setOf("tell_gudrun", "write_poem", "more_poem", "one_more_poem", "poem_done", "poem", "recital", "gunnars_ground")
```
GregHib: These can be moved outside of the function

GregHib:
Found dororans animation but still can't find gudruns for the life of me

jarryd229:
> Found dororans animation but still can't find gudruns for the life of me

dororans animation i think is a defense block animation with a sound don't think its right 

are they maybe render animations?
if not mite be hard to find i think its changed now for rs3


forgot to add advice() in Kjell.kts  fixed now

GregHib:
> dororans animation i think is a defense block animation with a sound don't think its right

Yeah you're right, removed.

> are they maybe render animations?

I had a look +/- 10 of their normal ones, doesn't look like it. Also checked models just incase it was an object.

I checked every dwarf animation according to animation skeletons, it doesn't exist as far as I can tell which is very odd considering the animation was there day 1 of the quests release.

GregHib (approved):
Cleaned up some of the cutscene, the scroll script fix requires the latest cache.

Issue #58, opened 2020-12-14
Simplified skill actions


GregHib:
Things to improve skilling grind and quality of life, typically options would change depending on if available or not e.g fire making "light" option only available if a tinderbox is in the inventory.

* "Light" option on logs for 1-click fire making
* "Offer" option on altars for 1-click bone offering
* Option order swapping (left click pickpocket etc..)
* "Make" options on herblore potions
* "Fletch" option on feathers/shafts/arrows/darts
* Auto theive thieving stalls
* Auto theive pickpocketting but really low xp, coins? Or leave pick pocketting for faster click intensive xp rates
* Withdraw-all from BoB familiar

Some of these might be too overpowered. Could also have one's which exist when near the object necessary, e.g:
* "Cook" raw foods when near a fire

I think making skills easier should only be paired with a low-ish general xp rate

GregHib:
RS3 also now has "Climb to top floor" options for places like lumbridge stairs

GregHib:
Click multiple to automatically clean herbs #481

GregHib:
Make-x up to 10 for bolt and arrow fletching #502

GregHib:
Osrs now often has skill repeat on an auto timer with a slower tick rate than doing manually

PR #481, opened 2024-03-14
Grimy herb cleaning for herblore

Syntax2022:
* Add Cleaning data class `Cleaning.kt`
* Add `HerbCleaning.kts` for herb cleaning logic
* Add `level requirement` `XP reward` on each grimy herb in `items.yml`

### DIFF
--- a/data/definitions/items.yml
+++ b/data/definitions/items.yml
@@ -864,6 +864,9 @@ grimy_guam:
   id: 199
   price: 140
   limit: 10000
+  cleaning:
+    level: 1
+    clean_xp: 2.5
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
   kept: "Wilderness"
@@ -872,6 +875,9 @@ grimy_marrentill:
   id: 201
   price: 29
   limit: 10000
+  cleaning:
+    level: 9
+    clean_xp: 5
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
   kept: "Wilderness"
@@ -880,6 +886,9 @@ grimy_tarromin:
   id: 203
   price: 307
   limit: 10000
+  cleaning:
+    level: 5
+    clean_xp: 3.8
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
   kept: "Wilderness"
@@ -888,6 +897,9 @@ grimy_harralander:
   id: 205
   price: 237
   limit: 10000
+  cleaning:
+    level: 20
+    clean_xp: 6.3
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
   kept: "Wilderness"
@@ -896,6 +908,9 @@ grimy_ranarr:
   id: 207
   price: 3423
   limit: 10000
+  cleaning:
+    level: 25
+    clean_xp: 7.5
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_ranarr_noted: 208
@@ -903,6 +918,9 @@ grimy_irit:
   id: 209
   price: 1877
   limit: 10000
+  cleaning:
+    level: 40
+    clean_xp: 8.8
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_irit_noted: 210
@@ -910,6 +928,9 @@ grimy_avantoe:
   id: 211
   price: 3886
   limit: 10000
+  cleaning:
+    level: 48
+    clean_xp: 10
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_avantoe_noted: 212
@@ -917,6 +938,9 @@ grimy_kwuarm:
   id: 213
   price: 2232
   limit: 10000
+  cleaning:
+    level: 54
+    clean_xp: 11.3
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_kwuarm_noted: 214
@@ -924,6 +948,9 @@ grimy_cadantine:
   id: 215
   price: 1184
   limit: 10000
+  cleaning:
+    level: 65
+    clean_xp: 12.5
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_cadantine_noted: 216
@@ -931,6 +958,9 @@ grimy_dwarf_weed:
   id: 217
   price: 4755
   limit: 10000
+  cleaning:
+    level: 70
+    clean_xp: 13.8
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_dwarf_weed_noted: 218
@@ -938,6 +968,9 @@ grimy_torstol:
   id: 219
   price: 43973
   limit: 10000
+  cleaning:
+    level: 75
+    clean_xp: 15
   weight: 0.007
   examine: "I need to clean this herb before I can use it."
 grimy_torstol_noted: 220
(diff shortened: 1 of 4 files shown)


Syntax2022:
Am i correct in thinking if i called `player.softTimers.start("timername")` it will automatically overwrite any existing timer because the name is the same or should i be calling 
```kotlin
if(player.softTimers.contains("timername")){
        player.softTimers.restart("timername")
} else {
        player.softTimers.start("timername")
}
```

GregHib:
Actually it'll just queue another timer with the same name, which would start calling `timerTick` twice a tick, which is not ideal and a bit of an oversight, I'll create an issue for it.

For context the timer's used in the other skills are mainly for the bots so that they can know when a skill they're performing has started/stopped/failed. In this case herbs take a fixed amount of time and never fail so don't worry about having such a timer.

game/src/main/kotlin/world/gregs/voidps/world/activity/skill/herblore/HerbCleaning.kts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+fun Player.clean(item: Item, player: Player, herb: Cleaning) {
+    if(!player.inventory.contains(item.id)) {
+        return
+    }
+    weakQueue("cleaning", 3) {
+        cleanHerb(player.inventory.indexOf(item.id), player, item, herb)
+        clean(item, player, herb)
+    }
```
GregHib: This doesn't seem correct, if you click one herb with more than one in the inventory it'll clean 2 of them?

I don't think there's a queue involved here as to my knowledge herblore can't be cancelled? Unless there's some mechanic I'm forgetting I think it just uses the skilling clock to prevent cleaning from happening too frequently

data/definitions/items.yml
```diff
@@ -1052,6 +1085,9 @@ clean_guam:
   id: 249
   price: 161
   limit: 10000
+  potion_unf:
+    level: 3
+    make: "guam_potion_unf"
```
GregHib: Left in by mistake?

GregHib (changes requested):

game/src/main/kotlin/world/gregs/voidps/world/activity/skill/herblore/HerbCleaning.kts
Syntax2022: > This doesn't seem correct, if you click one herb with more than one in the inventory it'll clean 2 of them?


No it cleans one as expected i will attach a GIF

cleaning


> I don't think there's a queue involved here as to my knowledge herblore can't be cancelled? Unless there's some mechanic I'm forgetting I think it just uses the skilling clock to prevent cleaning from happening too frequently

if i understand right what you're asking; in RuneScape when cleaning herbs you can click a single herb and it will automatically clean them with no interaction needed but you can speed cleaning up by clicking each herb in your inventory,  I am not not quite sure what you mean by it can't be cancelled? 

Video here: https://youtu.be/waww7COFxbU?si=yqk0TxLozG7R4o_t&t=15

data/definitions/items.yml
Syntax2022: Yes so sorry about that, I am also working on the rest of herblore and i thought i had removed this before pushing, would you like me to remove this?

GregHib: I presumed so, yeah might as well keep it separate

game/src/main/kotlin/world/gregs/voidps/world/activity/skill/herblore/HerbCleaning.kts
GregHib: Cancelled meaning you can interrupt and prevent the action from occuring part way through, like walking away when cooking or fishing. That's the use-case for queue's anyway.

Yeah so in that clip he's cleaning more than 1 herb per tick, so I don't think there's anything that needs doing other than just:

```kotlin
inventoryOption("Clean", "inventory") {
    val herb: Cleaning = item.def.getOrNull("cleaning") ?: return@inventoryOption
    if (!player.has(Skill.Herblore, herb.level, true)) {
        return@inventoryOption
    }
    
    if (player.inventory.replace(slot, item.id, item.id.replace("grimy", "clean"))) {
        player.experience.add(Skill.Herblore, herb.xp)
    }
}
````

Syntax2022: Okay no problem, The reason why i made it the way I did was because i was trying to reproduce this behaviour where it would clean them automatically after x ticks unless the player clicks the herb to clean them that's why i used the queue.  for example in osrs:

osrsexample

GregHib: Does it not work the same when you try ^ in-game on void? I would've thought it automatically occurs from the way packets are stored and processed at the start of each tick

Syntax2022: Just so we're both on the same page, because i think theres some confusion, when spam clicking the herb in Void it works as expected like OSRS, But in OSRS when i click the first herb i leave my mouse theres a timer that will clean the next herb in 2 ticks till all herbs are cleaned. if i were to replace the code i have with what you suggested it would removed the behaviour where if i leave my mouse it would clean the next herb in 2 ticks till they're all cleaned, sorry if i am not being clear.

GregHib: Okay so the first herb has a 2 tick delay and all ones after that are instant or also have 2 tick delay? You're right about the softQueue for the 2 ticks, what I didn't understand is why it was calling cleanHerb before the queue as well

Syntax2022: Okay so when you click the grimy herb it will clean that herb instantly then after that every 2 ticks it will clean the next herb that's why i call `cleanHerb(slot, player, item, herb)` to handle the instant cleaning then i call `player.clean(item, player, herb)` to handle the 2 tick delay cleaning

Heres a video i took with Void and Runescape side by side: https://streamable.com/ojsijm

GregHib: Thanks for the video it helped a lot. That's bizarre, that's not how I remember it working in rs2, osrs wiki has some mention of a herblore update but I can't find the exact mention of changes to the mechanics.

Syntax2022: You're right there were changes in September 20th 2020 is when they actually added the automatically cleaning mechanic in the update post is says "Clean Herbs; You may now clean all your herbs automatically after cleaning the first one in your Inventory." I completely missed this, and i always thought it was in RS2, would you like me to remove the automatically cleaning mechanic or keep it for QOL?

GregHib: I think let's leave it out for now, I've added it to #58 as it is a nice qol improvement but it could also be done with a more standard way with the make/clean x interface. I haven't really decided how closely to emulate vs enhance the gameplay so I've been making notes and putting stuff under the "beyond emulation" issue tag as things to do in the future once all the core content is complete

Syntax2022: No problem i have removed it, sorry about the confusion, I wasn't aware that was actually added 3 years ago, i thought it was a legacy mechanic

GregHib: No problem, I didn't even realise it was a mechanic in osrs, show's how often I play

GregHib (approved):

PR #515, opened 2024-04-11
Store charges in item amounts

### LINKED ISSUE #311: Item degrading & charges
Tracking items charges can be done on a per player or per item basis, ~~technically this era didn't have per-item tracking however it would be a nice to have instead of perfect emulation~~.
Rather than have `var charge: Int` in the Item class which complicates item containers, as for example stacking multiple items of the same type but with different charges (rs got around this by having those items not stackable in banks). It would be better to track items in a separate system.

```kotlin
val itemCharges: MutableMap<String, Int>

fun name(id: String, index: Int) = "${id}_$index"

on<ItemChanged> {
    val from = name(fromId, fromIndex)
    val charges = itemCharges.remove(from)
    if (charges != null {
        val to= name(toId, toIndex)
        itemCharges[to] = charges
    }
}

fun Container.charge(index: Int): Int {
    return itemCharges[name(id, index)] ?: 0
}
```
A base example. It would need to be extended to support a last-out stack for stacking non-stackable items in banks.

Items which use varps:
* Ring of recoil
* Black mask

Items which don't:
* dfs
* ancient warriors
* corrupt
* chaotics
* satchel
* tea flask

### LINKED ISSUE #514: Make Inventory operations more testable
Due to the unortherdox way operations are implemented by using interfaces to reduce boilerplate code, the classes and methods don't show up as not covered on tests (even though they absolutely are), giving `inv` 91% line but only 60% class and method coverage.

Could make operations more testable by changing them to classes or objects and have the methods extend TransactionOperation.

GregHib:
Closes #311

* Add way to revert non-item changes in transactions
* Remove item definitions storing from Item
* Use amount to store charges (as inventory level charges will always be 1 item and never stackable)
* Add add, remove and contains lists of item helpers
* Add more transaction and inventory unit tests
* Convert item operations to objects and extensions to make it more testable closes #514
* Verify unknown messages against osrs
* Rename ItemRemovalChecker to ItemAmountBounds


GregHib:
Thinking about it more I'm not sure this is the best approach, handling charges and amount separately is problematic and creates a divide and introduces complexity.

I think it might be better to instead add items with charges to the stackable list, and change removal with rules to be replace or remove with rules. That way all the code which applies to adding/removing/combining stacks get's reused for charges.

Then items which use player variables should be handled separately?

GregHib:
Having tried the other approach that also isn't suitable as you end up with a stackablility issue where when you want to add a new item to an inventory you don't want them to be combined into one stack but if you want to increase the charge you do, so you end up with this split in functionality anyway.

PR #529, opened 2024-05-01
Add light source crafting, lighting, extinguishing plus Giant Mole addition

Syntax2022:
* Add the ability to make lanterns and lamps to `item-on-item.yml`
* Add `LightSources` data class to handle when `lighting` and `extinguish` light sources
* Add `LightSource.kts` to handle the logic when `lighting` and `extinguish` light sources
* Add `Firemaking cape` and `Firemaking cape t` as accepted light sources for `Giant Mole`, including if the players are wearing these capes.

### DIFF
--- a/data/definitions/item-on-item.yml
+++ b/data/definitions/item-on-item.yml
@@ -6207,6 +6207,95 @@ rune_crossbow_finished:
     - item: rune_crossbow
   message: "You add a string to the stock."
   animation: fletching_string_rune_crossbow
+# light source making
+white_lantern_make:
+  skill: firemaking
+  level: 1
+  xp: 0.0
+  ticks: 1
+  remove:
+    - item: white_candle
+    - item: candle_lantern_empty
+  add:
+    - item: candle_lantern_white
+  message: "You add the candle to the empty lantern frame."
+black_lantern_make:
+  skill: firemaking
+  level: 1
+  xp: 0.0
+  ticks: 1
+  remove:
+    - item: black_candle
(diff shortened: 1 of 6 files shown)


Syntax2022:
Hey Greg,

With this i ran into some challenges when trying to distinguish different items because there wasn't a "universal" string i could use to ID every item used in this. for example ```itemOnItem("tinderbox*", "*candle*") {``` wouldn't work because theres, lanterns, a mining hat and a torch I am assuming not declaring an item when using `itemonitem` or `inventoryItem` would be bad practice.

is the current implementation fine due to the usage of the data class?

GregHib:
Yeah there's no super easy way, I've added [itemOnItems](https://github.com/GregHib/void/blob/9c4368f3ec239a62fd2980bce6e56f2fc180703b/engine/src/main/kotlin/world/gregs/voidps/engine/client/ui/interact/ItemOnItem.kt#L68) which takes two arrays if you want to pass in a list, otherwise it's just looping over `itemOnItem` or whatever the call is

Syntax2022:
> Yeah there's no super easy way, I've added [itemOnItems](https://github.com/GregHib/void/blob/9c4368f3ec239a62fd2980bce6e56f2fc180703b/engine/src/main/kotlin/world/gregs/voidps/engine/client/ui/interact/ItemOnItem.kt#L68) which takes two arrays if you want to pass in a list, otherwise it's just looping over `itemOnItem` or whatever the call is

i will do that, thank you

GregHib (approved):

PR #533, opened 2024-05-05
Overload fixes and revisions

Syntax2022:
I had noticed when making giant mole that the behaviour of overloads were slightly off, when applying the stats, it wasn't carrying over apon logging out, and the wilderness check would always return false.

For reference I have a video from 2009 and a forum post from 2010 for the mechanics that I changed

Video demonstrating when stats are boosted: https://youtu.be/7dmQtHVah80?si=o_PO0Emo0bF6nRfw&t=43 

Forum post stating that potions don't reset when logging out, reply by user ```Zaaps1```: 'https://forum.tip.it/topic/254557-if-i-log-out-will-my-super-anti-posion-stop-its-effect/'


game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
```diff
@@ -66,15 +74,20 @@ timerTick("overload") { player ->
     }
 }
 
-timerStop("overload") { player ->
-    reset(player, Skill.Attack)
-    reset(player, Skill.Strength)
-    reset(player, Skill.Defence)
-    reset(player, Skill.Magic)
-    reset(player, Skill.Ranged)
+fun removeBoost(player: Player) {
```
GregHib: Probably needs to remove boost when entering wilderness too?

Syntax2022:
Just to give more clarification on why i removed the ability to use overloads in the wilderness, it's because it was only after EOC where overloads were allowed to be used in the wilderness.

GregHib:
Yeah absolutely they are far too OP to be allowed in the wilderness

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
```diff
@@ -11,17 +13,32 @@ import world.gregs.voidps.engine.timer.timerStart
...
         cancel()
+        return@canConsume
     } else if (player.levels.get(Skill.Constitution) < 500) {
         player.message("You need more than 500 life points to survive the power of overload.")
         cancel()
+        return@canConsume
+    }
+}
+
+enterArea("wilderness*") {
```
GregHib: It should just be `wilderness` the others are multi areas which can extend outside of the actual pvp area, see; map viewer
image

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
```diff
@@ -11,17 +13,32 @@ import world.gregs.voidps.engine.timer.timerStart
...
+    if (player.timers.contains("overload")) {
+        player.timers.stop("overload")
+        player["overload_refreshes_remaining"] = 0
```
GregHib: I don't think we should be removing it?

Aka if a player is fighting kbd, exits through the wilderness, then re-enters I imagine they'd lose their current stats, but once the timer ticks it'd get re-boosted so long as they're in kbd area?

Syntax2022: Yeah you're right i am just fixing that now

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
Syntax2022: I am unable to get the map viewer to work properly for me. what i see is a black screen with the areas outlined i will attached a screenshot. @GregHib (sorry for the ping but just wanted to make sure you see this)
map

jarryd229: under world.gregs.voidps.tools.map.render you need to run WorldMapDumper.kt to dump the images then run map viewer

Syntax2022: > under world.gregs.voidps.tools.map.render you need to run WorldMapDumper.kt to dump the images then run map viewer

Thank you, didn't know i had to run that before running the viewer

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
```diff
@@ -11,17 +14,36 @@ import world.gregs.voidps.engine.timer.timerStart
...
+    if (player.timers.contains("overload")) {
+        removeBoost(player)
+    }
+}
+
+exitArea("wilderness") {
+    if (player.timers.contains("overload")) {
+        applyBoost(player)
     }
 }
```
GregHib:
```suggestion
exitArea("wilderness") {
    if (player.timers.contains("overload")) {
        applyBoost(player)
    }
}
```

Should be left to the normal timer to boost, this could in theory be abused by re-applying the boost faster by quickly enter/exiting the wilderness

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
```diff
@@ -49,32 +72,37 @@ timerStart("overload") { player ->
 timerTick("overload") { player ->
     if (player.dec("overload_refreshes_remaining") <= 0) {
         cancel()
+        removeBoost(player)
```
GregHib: This should be in `timerStop`, in-case something else cancels the timer (command, death, etc..)

Syntax2022: Okay what I noticed, timer stop is called when you log out, your boosted stats aren't saved, do you think it'll be sufficient to removed boost on timer stop and apply boost on timer restart?

GregHib: Hmm it should save stats on logout, that might be a bug

Syntax2022: let me double check and i will update you here. just want to make sure i am right and i didn't do anything wrong.

Syntax2022: > Hmm it should save stats on logout, that might be a bug

Just double checked and the boost is removed but the variable is stored in the player save

GregHib: Ah okay it's because timerStop is being called on logout, so it's removing the boost and the remaining refreshes variable, which is not correct.

```kotlin
timerStop("overload") { player ->
    if (logout) {
        return@timerStop
    }
    //...
}
```
Should fix it

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
Syntax2022: Can you elaborate why applying the boost instantly would be a bad thing, my thought process is if the boost can't be used during PVP applying the boost instantly shouldn't be an issue. also, when drinking the overload the boost is applied instantly.

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
Syntax2022: @GregHib i see you said " so it's removing the boost and the remaining refreshes variable" when the timer is stopped during logout it doesn't remove the remaining refreshes from the player, it still persists over logging. i have attached a gif, this might be a bug or i am misunderstanding what you're referring too.

Peek 2024-05-13 17-16f

GregHib: I was mainly referring to your levels which are being reset to 99, but it was also true for the refreshes remaining variable, but looks like you've removed that from the timerStop code, [it was there before](https://github.com/GregHib/void/blob/142a65daf4719cb1e8c1332303059e15eb460565/game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts#L77).

It should still be there as a precaution

game/src/main/kotlin/world/gregs/voidps/world/interact/entity/player/combat/consume/drink/Overload.kts
GregHib: Sorry for the slow responses, if your stats are drained after you have overloaded e.g. because of kbd's "shock" firebreath. The overload would boost it back up again after 15 seconds, but by enter/exit the wildy you could get it boosted back-up sooner than you should be allowed to. It's not really a big deal tbh, just a small detail.

GregHib (approved):
I quickly finished it off so I can get it merged for the next update :)

PR #554, opened 2024-09-07
druidic ritual quest

jarryd229:
* Add Druidic Ritual quest
* Fix Gunnars Ground Gudrun hug animation
* Add women to lumbridge
* Fix walking diagonally through double doors
* Convert tollgate to use Doors.enter

### DIFF
--- a/data/definitions/animations.yml
+++ b/data/definitions/animations.yml
@@ -1799,3 +1799,4 @@ wolves_attack: 6559
 wolves_death: 6558
 run_replenish: 9988
 play_organ: 3675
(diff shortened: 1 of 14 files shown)


GregHib (approved):
Thanks, looks good, I just tidied up and added the suit of armour doors in the dungeon. I looped through ~2000 human animations and couldn't find the [armour stand animation seen here](https://youtu.be/AYGRfwQbuGE?si=D11lNv3FXHpIAGa_&t=168) however osrs doesn't have one so it's not the end of the world.

PR #615, opened 2025-02-20
Add ape atoll agility course

GregHib:
- Add monkey greegrees
- Add ape atoll agility course + tests
- Fix sending empty delta's to exact move
- Fix render emote reset ignoring transforms


game/src/main/kotlin/content/quest/member/monkey_madness/Greegrees.kts
Hc747, line 1: This plugin/scripting system is gorgeous

PR #622, opened 2025-03-17
Add lazy loaded CRC keys to cache interface for use in login protocol

titandino:
Feel free to close if it's already implemented somewhere or not planned on being supported, but I didn't see any place where the client's CRCs are validated on login anywhere. Does 667 just not send these values or are they currently just being skipped on the login protocol?

### DIFF
--- a/cache/src/main/kotlin/world/gregs/voidps/cache/Cache.kt
+++ b/cache/src/main/kotlin/world/gregs/voidps/cache/Cache.kt
@@ -10,6 +10,8 @@ interface Cache {
 
     fun indices(): IntArray
 
+    fun indexCrcs(): IntArray
+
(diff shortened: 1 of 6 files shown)


GregHib:
Hey, yeah Void is skipping the lobby login using [two](https://github.com/GregHib/void-client/blob/72fabcdc715b063a4fc7a22d1b66d811898a7ed9/client/src/Class14_Sub4.java#L216) client [changes](https://github.com/GregHib/void-client/blob/025e51d0f540fc9c9ebb50fd42ac4507fe43c423/client/src/Class151.java#L75) and so doesn't need CRCs as they're only part of the lobby login protocol.

titandino:
Gotcha gotcha!

PR #661, opened 2025-05-04
Fix Equip Requirements

lucid-soft:
Equip should requirements always check max level instead of boostable level, to prevent being able to stat boost to equip items they shouldn't be able to (e.g. Using a defence potion at lvl 1 defence to be able to wear steel armour)


GregHib (approved):
Thanks for the fix! Couldn't tell you why this is the way it is 😅

PR #676, opened 2025-05-09
Fix: Resolve Account Lookup Issues on Nix Environments and Enforce Chat Capitalization Rules

HarleyGilpin:
### 1. **Case Sensitivity in Account Lookups**
Account lookup failures were occurring on Linux systems due to case-sensitive string comparisons of usernames. This discrepancy stems from differences in how Linux and Windows handle file and string casing.

- ✅ **Fix:** Usernames are now normalized to lowercase before performing account lookups, ensuring consistent behavior across platforms.

### 2. **Public Chat Capitalization and Anti-Caps Spam**
To improve chat readability and mimic RuneScape’s formatting conventions:

- Messages that are fully capitalized (e.g., `HELP ME`) are automatically reformatted so that only the first letter of each word is capitalized (`Help Me`).
- The first word of every sentence is now capitalized to enforce basic grammar structure and deter caps lock spam.

### DIFF
--- a/data/area/asgarnia/burthorpe/burthorpe.teles.toml
+++ b/data/area/asgarnia/burthorpe/burthorpe.teles.toml
@@ -4,51 +4,6 @@ option = "Climb-down"
 tile = { x = 2927, y = 3444 }
 to = { x = 2209, y = 5348 }
 
-[1738]
-option = "Climb-up"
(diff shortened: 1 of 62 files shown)


GregHib (approved):
Looking good, thanks!

PR #677, opened 2025-05-11
Add playlist functionality

Ilwyd:
- Implemented the song end packet / event
- Added functionality to the music player playlist


engine/src/main/kotlin/world/gregs/voidps/engine/client/ui/SongEndEvent.kt
```diff
@@ -0,0 +1,24 @@
...
+
+data class SongEndEvent(
+    val songIndex: Int
+) : Event {
+    override val size = 2
+
+    override fun parameter(dispatcher: EventDispatcher, index: Int): Any? {
+        return when(index) {
+            0 -> "song_end"
+            1 -> songIndex
```
GregHib, line 16: Doesn't need to be changed but as the handler isn't matching against songIndex, index 1 isn't needed here

GregHib (approved):
Great work!

game/src/main/kotlin/content/entity/world/music/Music.kts
```diff
@@ -62,11 +92,128 @@ move({ !it.isBot }) { player ->
...
+    player["playlist_slot_${fromSlot+1}"] = toSong
+    player["playlist_slot_${toSlot+1}"] = fromSong
+}
+
+songEnd {player ->
+    player["playing_song"] = false
+    if (player["playlist_enabled", false]) {
+        playNextPlaylistTrack(player, songIndex)
+        return@songEnd
+    }
```
GregHib:
```suggestion
    if (player["playlist_enabled", false] && playNextPlaylistTrack(player, songIndex)) {
        return@songEnd
    }
```

If a playlist is cleared `playlist_enabled` will still be true but there won't be any tracks, so it would be a good idea to default to playing the area track (or alternatively setting playlist_enabled = false when clearing a playlist?)

game/src/main/kotlin/content/entity/world/music/Music.kts
```diff
@@ -43,11 +46,38 @@ fun playAreaTrack(player: Player) {
     }
 }
 
+fun playNextPlaylistTrack(player: Player, finishedTrackId: Int) {
+    val finishedTrackIndex = enums.get("music_tracks").getKey(finishedTrackId)
+    val playlistTracks = (1..12).map { player["playlist_slot_$it", 32767] }.filter { it != 32767 }
+
+    if (playlistTracks.isEmpty()) return
```
GregHib:
```suggestion
    if (playlistTracks.isEmpty()) return false
```

GregHib (changes requested):
Noticed a small issue

game/src/main/kotlin/content/entity/world/music/Music.kts
Ilwyd: Good catch. Thanks, resolved

GregHib (approved):

PR #721, opened 2025-06-15
Implemented Imp with teleport, drop table, sounds and animations

M1krofin:
Implemented Imp with teleport, drop table, sounds and animations

Issue #503, opened 2024-03-27
Slayer


GregHib:
- [x] Slayer masters
  - [x] Turael
  - [x] Mazchna
  - [x] Vannaka
  - [x] Chaeldar
  - [x] Sumona
  - [x] Duradel
  - [x] Kuradal
- [ ] Slayer equipment
  - [x] Enchanted gem
  - [x] Black mask
  - [x] Slayer helmet
  - [x] Spiny helmet
  - [x] Facemask
  - [x] Earmuffs
  - [x] Bag of salt
  - [x] Ice cooler
  - [x] Mirror shield
  - [x] Unlit bug lantern
  - [x] Witchwood icon
  - [x] Insulated boots
  - [ ] Slayer bell
  - [x] Slayer gloves
  - [x] Leaf-bladed weapons
  - [x] Slayer's staff
  - [x] Broad equipment
  - [ ] Crystal chime
  - [x] Fungicide
  - [x] Nose peg
  - [x] Rock hammer
- [x] Points/rewards
- [ ] Slayer monsters
  - [x] Slayer Tower
  - [x] Fremennik Slayer Dungeon
  - [x] Lumbridge Swamp Caves
  - [x] Kalphite Cave
  - [x] Kuradals Dungeon
  - [x] Pollnivneach Slayer Dungeon
  - [ ] Poison waste dungeon
  - [x] Chaos dungeon
  - [ ] Guthixian Temple
  - [x] Zanaris

Depends on #378

GregHib:
Monsters not added:
- [ ] warped creatures
- [ ] elves
- [ ] scabarites
- [x] living rock creatures
- [ ] terror dogs
- [x] tormented demons
- [ ] mogres
- [x] shades
- [ ] sea_snakes

GregHib:
All the important ones done now. The rest can be left for their respective quests and elves can wait for the isafdar forest shortcuts.

PR #725, opened 2025-06-24
Add slayer tower

GregHib:
#503 
- Add npc spawns, definitions and drops by @MrSlayerGod 
- Add stairs, chain shortcuts, doors
- Add abyssal demons, nechryaels, aberrant spectres
- Tidy up spawns and monster config files
- Fix teleporting and walking npc updating as seen in #721


GregHib:
Copy of #712 without all of the merge chaos which was causing issues

PR #739, opened 2025-08-07
Fix ground items not showing when changing levels

### LINKED ISSUE #738: Items do not spawn on other floors when logged in first time after starting server
**Describe the bug**
Items in different floors does not spawn normally after logged in first time

**To Reproduce**
1. Start server and login to account on first floor 
2. Go to the second or third floor on lumbridge castle
3. See that no logs spawn
4. log out and log back in
5. logs have spawned.

**Expected behavior**
I should see the logs spawned or dagger. Same goes to the bronze pickaxe spawn 

**Screenshots**

https://github.com/user-attachments/assets/96a57fde-7165-44cd-9d04-b25be35dc502

**Desktop (please complete the following information):**
Newest build

Ilwyd:
Fixes #738

Items are now shown to the player when they change floors.

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/client/update/batch/ZoneBatchUpdates.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/client/update/batch/ZoneBatchUpdates.kt
@@ -49,7 +49,7 @@ class ZoneBatchUpdates : Runnable {
 
     fun run(player: Player) {
         val previousZone: Zone? = player["previous_zone"]
-        val previous = previousZone?.toRectangle(radius = player.viewport!!.localRadius)?.toZones(player.tile.level)?.toSet()
+        val previous = previousZone?.toRectangle(radius = player.viewport!!.localRadius)?.toZones(previousZone.level)?.toSet()
         player["previous_zone"] = player.tile.zone
         for (zone in player.tile.zone.toRectangle(radius = player.viewport!!.localRadius).toZonesReversed(player.tile.level)) {
             val entered = previous == null || !previous.contains(zone)


engine/src/main/kotlin/world/gregs/voidps/engine/client/update/batch/ZoneBatchUpdates.kt
```diff
@@ -49,10 +49,12 @@ class ZoneBatchUpdates : Runnable {
 
     fun run(player: Player) {
         val previousZone: Zone? = player["previous_zone"]
+        val previousLevel: Int? = player["previous_level"]
         val previous = previousZone?.toRectangle(radius = player.viewport!!.localRadius)?.toZones(player.tile.level)?.toSet()
```
GregHib: Why this works didn't make sense to me because `Zone` already includes a `level` so I had a look.
Turns out because `toZones(player.tile.level)` is using the current level, rather than the previous it means `previous` always contains the current zone unless teleporting.

So you can revert your changes and just change this to fix

```suggestion
        val previous = previousZone?.toRectangle(radius = player.viewport!!.localRadius)?.toZones(previousZone.level)?.toSet()
```

GregHib (changes requested):

engine/src/main/kotlin/world/gregs/voidps/engine/client/update/batch/ZoneBatchUpdates.kt
Ilwyd: Ahh, good catch. I'll make that change :)

GregHib (approved):

PR #740, opened 2025-08-11
Add Summoning followers

Ilwyd:
This is a WIP PR to add summoning followers to the game. This is intended to just the be basics of having followers and then in a future PR things like familiar combat, scroll effects, etc can be added.

- [x] Using pouches to spawn familiars
- [ ] ~Dropping / spawning pets~
- [x] Familiar details tab
  - [x] Familiar chathead
  - [x] Familiar name
  - [x] Correct chathead animations
  - [x] Familiar timer
    - [x] Dismiss familiar on time running out
  - [x] Call
  - [x] Dismiss
  - [x] Renew familiar
- [ ] ~Pet details tab~
  - [ ] ~Pet chathead~
  - [ ] ~Pet name~
  - [ ] ~Correct chathead animations~
  - [ ] ~Pet hunger (maybe)~
  - [ ] ~Pet age (maybe)~
  - [ ] ~Dismiss, call pet buttons~
- [x] Summoning orb right-click menu
  - [x] Select left-click option menu
  - [x] Dismiss
  - [x] Call
  - [x] Renew
- [x] Saving follower details on logout
- [x] Spawning follower again on login

Known Bugs:
- [x] The cast option for every familiar you've summoned in your session shows in the right click menu for the summoning orb
- [x] Logging out with a familiar breaks the summoning orb on summoning a new familiar on your next log in


engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/player/Player.kt
```diff
@@ -47,6 +48,7 @@ class Player(
     var accountName: String = "",
     var passwordHash: String = "",
     val body: BodyParts = BodyParts(),
+    var follower: NPC? = null,
```
GregHib: Variables can be used for this, a custom var storing the follower id will be needed, it's also better to store the npc index and look-up the NPC rather than store the NPC directly to prevent accidentally using (and the garbage collector not being able to delete) npcs that no-longer exist
e.g.

```kotlin
var Player.follower: NPC?
    get() {
        val index = this["follower_index", -1]
        get<NPCS>.get(index)
    }
    set(value) {
        this["follower_index"] = value.index
    }
```

Ilwyd: Thanks, I was planning on changing this. Wrote that quickly just so I could get the dismiss button working.

One problem I'm having with this approach is that index is -1 for a tick, so I can't immediately set follower when the npc is added. I can queue it, not sure if that's the correct approach though.

GregHib: Hmm yeah that's a good point, using the world queue for that should be fine just make sure the name is unique

DavidMBusey:
This is sick.

Ilwyd:
> This is sick.

Thanks, slowly working my way through it when I have the time 😅

Ilwyd:
Screwed up my rebase I guess. Will fix it

Ilwyd:
Branch should be all fixed up now

GregHib (approved):
Fixed up some little bits to round off what was worked on so it can be merged in

Ilwyd:
> Fixed up some little bits to round off what was worked on so it can be merged in

Thanks for getting this in (and sorry for abandoning it :( )

GregHib:
> sorry for abandoning it

Haha dw even almost complete contributions help :smile:

Issue #746, opened 2025-08-18
Add missing tests for quests


GregHib:
- [x] Cooks assistant
- [x] Demon slayer
- [x] Dorics quest
- [x] Gunnars ground
- [x] Restless ghost
- [x] Rune mysteries
- [x] The knights sword
- [x] Prince Ali Rescue
- [x] Enter the abyss
- [x] Druidic Ritual

Full coverage not needed, just the happy path to complete the quest.

PR #816, opened 2025-12-21
Test the happy path of the Cook's Assistant quest

dragbone:
What better way to discover and understand a code base than to write a test 😉
Should resolve one task of #746


GregHib (approved):
Thanks for the contribution it absolutely is a great way to learn it! The dialogue fast forward is a nice touch I'm going to consider using that in other quest tests as skipping exact counts doesn't actually prove anything it just mains test maintenance harder if dialogues change

Issue #830, opened 2026-01-15
Ruby bolts (e) don't seem to proc frequently enough


GregHib:

Syntax2022:
hey Greg, hope you’re doing well  I looked into the code and it seems the Enchanted bolt effects are currently gated by baseDamage < 0.

Expected behavior should be that bolt effects can still proc chance even on a 0 hit.

see here [https://oldschool.runescape.wiki/w/Ruby_bolts_(e)#Special_effect](url)

PR #841, opened 2026-01-19
Fix enchanted bolts

Syntax2022:
- Add rollOnMissEnchantments and lets those enchanted bolts roll their special even on misses. #830

### DIFF
--- a/game/src/main/kotlin/content/skill/ranged/Ammo.kt
+++ b/game/src/main/kotlin/content/skill/ranged/Ammo.kt
@@ -31,19 +31,6 @@ object Ammo {
 
     fun requiredAmount(weapon: Item, special: Boolean) = if (weapon.id.startsWith("dark_bow") || (weapon.id.startsWith("magic_shortbow") && special)) 2 else 1
 
-    private val rollOnMissEnchantments = setOf(
-        "opal_bolts_e",
-        "jade_bolts_e",
-        "pearl_bolts_e",
-        "topaz_bolts_e",
-        "sapphire_bolts_e",
-        "emerald_bolts_e",
-        "ruby_bolts_e",
-        "diamond_bolts_e",
-        "dragon_bolts_e",
-        "onyx_bolts_e",
-    )
-
     fun remove(player: Player, target: Character, ammo: String, required: Int) {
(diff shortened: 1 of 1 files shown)


game/src/main/kotlin/content/skill/ranged/Ammo.kt
```diff
@@ -76,10 +89,15 @@ object Ammo {
...
         baseDamage: Int,
     ): Int {
-        if (source !is Player || baseDamage < 0 || type != "range") {
+        if (source !is Player || type != "range") {
+            return baseDamage
+        }
+        val miss = baseDamage < 0
+        if (miss && source.ammo !in rollOnMissEnchantments) {
             return baseDamage
         }
```
GregHib: This can simply be removed as it's a method only for enchanted bolts
```suggestion
```

game/src/main/kotlin/content/skill/ranged/Ammo.kt
```diff
@@ -31,6 +31,19 @@ object Ammo {
...
+        "pearl_bolts_e",
+        "topaz_bolts_e",
+        "sapphire_bolts_e",
+        "emerald_bolts_e",
+        "ruby_bolts_e",
+        "diamond_bolts_e",
+        "dragon_bolts_e",
+        "onyx_bolts_e",
+    )
+
```
GregHib:
```suggestion
```

GregHib (changes requested):
Thanks for looking into it, quick tweak

game/src/main/kotlin/content/skill/ranged/Ammo.kt
Syntax2022: Okay i'll change that, the reason i had done it that way, was just in case in the future, there would be some sort of bolt that would be affected by hit chance, that allows for both scenarios.

GregHib (approved):

game/src/main/kotlin/content/skill/ranged/Ammo.kt
GregHib: It won't apply to other types of bolts as each check starts with `source.ammo == ` everything which isn't enchanted get's skipped over

PR #858, opened 2026-01-25
Add shop definition for Nurmofs Pickaxe Shop

cadyyan:
The ID of the shop was pulled from this thread: https://rune-server.org/threads/all-2011-shops.693716/
The stock of items was pulled from this wiki page: https://oldschool.runescape.wiki/w/Nurmof

Tested in a private server.


GregHib (approved):
Thanks!

PR #879, opened 2026-02-04
Fix teleport level bounds for rune essence mine

cadyyan:
The rune essence mine area didn't have any level bounds defined which meant that when the player was teleported they could land on levels 0-3 randomly. Unfortunately only level 0 is valid which would leave the player stuck unless they could teleport out using a spell (the portals are on level 0).

To fix this I've added a max level of 0 to the area definition.

To confirm this fix I teleported using Aubury dozens of times.


GregHib (approved):
Thanks, definitely due to an inconsistency between Polygons and Rectangles and how random values within them are chosen.

PR #880, opened 2026-02-05
Add "pot of flour" to Aggie skin paste dialogue

cadyyan:
While attempting to complete the Prince Ali Rescue quest I got stuck on the skin paste task. The dialogue said you only need 3 items but trying to give them to Aggie doesn't do anything. Checking the actual implementation and the quest wiki shows that you also need a "pot of flour" to make the skin paste.

I've updated the text to match that.


GregHib (approved):
Nice find, thanks

PR #881, opened 2026-02-05
Add Ninja+kotlinc build with offline vendor support

tritao:
## Summary
Adds a non-Gradle build/run workflow using Ninja + `kotlinc`, plus an optional vendored-deps path for offline builds.

## What’s included
- `Makefile`: default `make build/run` uses Ninja+`kotlinc` (module-level incremental rebuilds).
- Repo-local bootstraps:
  - Temurin JDK 21 (`make bootstrap`)
  - Kotlin compiler (`make kt-bootstrap-kotlin`)
  - `coursier` (`make kt-bootstrap-cs`)
- Cache helper: `make cache-bootstrap` populates `./data/cache` from common local locations (`data/cache.zip`, newest `cache/*.7z`, etc), or `CACHE_ZIP/CACHE_DIR/CACHE_URL`.
- Offline deps:
  - `make vendor-deps` vendors resolved Maven jars into `./vendor/`
  - `make vendor-verify` checks `vendor/SHA256SUMS`
  - `KT_OFFLINE=auto|0|1` (default `auto`: prefer `vendor/` if present, otherwise download)
- Gradle workflow kept available via `gradle-*` targets.

## How to use
- Online (no vendoring required): `make run`
- Offline (no downloads): `make vendor-deps` (once, online) then `make run` (uses vendor automatically) or `make run KT_OFFLINE=1`.

## Notes for reviewers
- This PR includes vendored jars under `vendor/` (~tens of MB) to enable fully offline builds. If you’d rather not keep binaries in git, I can split this PR or drop the vendoring commit.
- Cache files are *not* included; they’re still external and expected under `./data/cache/`.

## Validation
- `make build`
- `make build KT_OFFLINE=1`
- `make vendor-verify`


GregHib:
Definitely going to need a compelling reason to allow this in, what benefits is ninja bringing that gradle and it's wrapper doesn't already do, especially given the downsides of the jars and python noise, if it's speed I'm going to need to some real numbers.

tritao:
> Definitely going to need a compelling reason to allow this in, what benefits is ninja bringing that gradle and it's wrapper doesn't already do, especially given the downsides of the jars and python noise, if it's speed I'm going to need to some real numbers.

Agreed.

I got some numbers yesterday ninja is slower with a clean build, but a bit faster for no-op/incremental builds.

My main reason for this change was actually to be able to have hermetic builds, that is, I should be able to build the repository without network acess, as well as 10 years from now once all the Java library asset servers are dead.

I've had similar build issues with older projects, so it's not just a theoretical issue.

If you think that's valuable, and don't mind the vendored the dependencies, I can check if its possible to achieve it with Gradle too.

GregHib:
> I can check if its possible to achieve it with Gradle too

Yeah gradle has an offline mode which you can toggle (once it's downloaded all the dependencies once) and you can achieve the same effect by caching the .gradle directory - which is what the github actions CI does so it doesn't have to re-download dependencies each run.

> as well as 10 years from now once all the Java library asset servers are dead.
> I've had similar build issues with older projects, so it's not just a theoretical issue.

It is somewhat hypothetical. The only large maven service removed in recent years was JCenter/Bintray in 2021 and they gave people a year's warning to migrate off of the platform. For an actively maintained repository this doesn't cause any issues.

And even if a repository is no longer maintained and falls victim to such a shutdown the dependencies are still likely to exist elsewhere and can be easily replaced. Even in the worse case a dependency was wiped clean from the internet; it'll still exist and could be extracted from the fat-jars which are in [releases](https://github.com/GregHib/void/releases).

For concerns of longevity it would be much simpler to archive those ~25mb zip files than to introduce a secondary build system.

> I got some numbers yesterday ninja is slower with a clean build, but a bit faster for no-op/incremental builds.

Yeah it doesn't seem worth the trade-off of having to concurrently maintain two build systems, especially one that I'm not familiar with and isn't standard in the java industry, I'll be closing this but am happy to continue discussing if necessary.

PR #882, opened 2026-02-06
Add direction to Musa Point fishing spots

cadyyan:
The Musa Point fishing spots didn't actually work when you tried to fish at them. When you tried to use a small net, cage, or harpoon to fish, nothing happens and there is nothing in the chat to tell you that something is wrong.

Comparing the fishing spots to other spots that I know work I found that the direction component of the NPC spawn was missing here. Figured out the directions in game and added the directions fixed the fishing spots.


GregHib:
Thanks, looks like this is due to branch being out-dated and the spawn points not having any fishing data on them, rather than due to not being able to reach them.

Update to the latest main branch and this should work fine now.

cadyyan:
Ah yeah you're right. I thought I had pulled the latest changes in before building but must have missed that!

GregHib:
I only realised this was the cause because I did the same thing 😆

PR #890, opened 2026-02-13
Add ability to create talisman tiaras

cadyyan:
Talisman tiaras had their general mechanics implemented already but you couldn't actually craft them at the altar yet.

This adds the ability to create talisman tiaras by taking a blank tiara and a talisman to a rune altar and using the tiara on the altar. Added an automated test for this behavior as well as the requisite experience configuration to the TOML files. Experience per craft were taken from the OSRS wiki.

### DIFF
--- a/data/skill/runecrafting/runecrafting.items.toml
+++ b/data/skill/runecrafting/runecrafting.items.toml
@@ -182,6 +182,7 @@ limit = 100
 weight = 1.0
 slot = "Hat"
 type = "HairMid"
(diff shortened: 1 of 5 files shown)


cadyyan:
Weird... I'm not seeing that error when running tests on this branch locally....

GregHib:
> Weird... I'm not seeing that error when running tests on this branch locally....

Dw it was an error with some outdated CI nothing to do with your code, I fixed in #891 and have rebased so it's now passing

PR #896, opened 2026-02-18
Add behavior for Antique Lamp from Gunnars Ground

cadyyan:
Currently if you complete the quest you're awarded the lamp but rubbing it does nothing and you can't bank the item. The only option is to destroy it.

The wiki says that the expected behavior is to be able to select a skill of at least level 5 and you'll be granted 200 experience.


GregHib (approved):
Love to see these little qol improvements

PR #900, opened 2026-02-26
Fixed bugs with the Delrith encounter in Demon Slayer

cadyyan:
While completing the Demon Slayer quests I found few bugs that
would be confusing for players.

1. If you give the wrong word for the incantation Delrith just
   disappears and doesn't come back. You could work around this by
   walking out of the area which brings you out of the instance and then
   re-entering the instance. This is a bit annoying and not what happens
   normally. The change here was to reset Delrith back to full strength
   and have him attack the player again.
2. If the player logs out while in the instance, when they log back in
   they'll be floating in an empty instance. The only way out is to
   teleport somewhere which could require the player to wait a while if
   the only teleport they have is the base Lumbridge one and they used
   it recently. The fix here was just to catch the logout event and
   teleport them to safety before tearing the instance down.
3. I noticed that when you got to actually fighting Delrith, after slaying him
   if you clicked the Banish option you would get a "Nothing interesting
   happens" message. Looks like the expected target was set to full strength
   Delrith on accident. This had no workaround for players and needed a fix.

### DIFF
--- a/data/entity/npc/monster/demon/demons.npcs.toml
+++ b/data/entity/npc/monster/demon/demons.npcs.toml
@@ -6,6 +6,7 @@ str = 10
 def = 5
 combat_def = "delrith"
 categories = ["demons"]
+hunt_mode = "aggressive"
 examine = "A freshly summoned demon."
 
 [delrith_weakened]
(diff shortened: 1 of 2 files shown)


GregHib:
Hey thanks for this, I won't be merging because I've done a more fundamental fix for 2. in #906.

Nothing happening after failing Delriths incantation is correct; albeit not player friendly, he's also not supposed to be aggressive last I checked. Your fix for interacting with the weakened delrith was correct though, nice find.

PR #949, opened 2026-04-13
Add Clan Wars Free-For-All (Safe & Dangerous) portals     

HarleyGilpin:
## Add Clan Wars Free-For-All (Safe & Dangerous) portals                                                                 
                                                                                                                                                                                        
  ### Summary                                                                                                                                                                           
  - Implement entry and exit portals for both FFA Safe and FFA Dangerous arenas                                                                                                         
  - Both portals share the same warning interface (id 793), with varbit 5279 selecting which text the client displays                                                                   
  - Warning dismiss preferences are stored independently per portal and integrate with the Doomsayer system                                                                             
  - Safe arena keeps items on death, dangerous arena drops items on death                                                                                                               
  - PvP is enabled only inside the fighting areas, with the Clan Wars overlay shown across both arenas                                                                                  
  - Challenge portal (purple) shows an "under construction" message                                                                                                                     
  - Tested in-game end-to-end
  
   ### Bug fixes & Improvements
   - Dragon claws special attack.
   - Dark bow can now shoot dragon arrows.
   - Fixed game crash in the Fight Caves minigame.
   - Fixed a bug with character account creation
   - Bank tab creation via drag-and-drop was silently ignored because the interfaceSwap handler only matched on fromId, causing the generic inventory swap handler to intercept the event instead.                                                                                                                                                                                                                                                                                        
   - Collapsing a tab with more than two rows of items crashed the client because shiftToFreeIndex generated O(N*M) intermediate inventory updates in a single packet. Replaced with a single-pass bulk rearrangement.      
 
  
[Screencast_20260412_143216.webm](https://github.com/user-attachments/assets/2ad47ad8-e3ac-45a5-9f0c-b853a515bb76)


game/src/main/kotlin/content/minigame/clan_wars/ClanWarsFreeForAll.kt
```diff
@@ -0,0 +1,131 @@
...
+        objectOperate("Enter", "clan_wars_portal_ffa_safe") {
+            if (combatLevel < 30) {
+                message("You need a combat level of at least 30 to enter this portal.")
+                return@objectOperate
+            }
+            set("clan_wars_ffa_portal", 0)
+            if (get("warning_clan_wars_ffa_safe", 0) == 7) {
+                tele(safeArena)
+                return@objectOperate
+            }
```
GregHib: You should be able to reuse the warning method here and not need to re-implement all the interface options

```suggestion
            if (warning("clan_wars_ffa_safe")) {
                tele(safeArena)
                return@objectOperate
            }
```

See Warning.kt

game/src/main/kotlin/content/minigame/clan_wars/ClanWarsFreeForAll.kt
```diff
@@ -0,0 +1,131 @@
...
+        playerSpawn {
+            if (tile in Areas["clan_wars_ffa"]) {
+                open("clan_wars")
+                if (tile in Areas["clan_wars_ffa_safe_arena"] || tile in Areas["clan_wars_ffa_dangerous_arena"]) {
+                    set("in_pvp", true)
+                    options.set(1, "Attack")
+                }
+            }
+        }
+    }
```
GregHib: Not necessary, `entered("clan_wars_ffa")` and `entered("clan_wars_ffa_dangerous_arena")` will be triggered on login too
```suggestion
```

game/src/main/kotlin/content/skill/melee/weapon/special/DragonClaws.kt
```diff
@@ -14,32 +15,39 @@ class DragonClaws : Script {
...
             val weapon = weapon
             var (hit1, hit2, hit3, hit4) = intArrayOf(0, 0, 0, 0)
             val maxHit = Damage.maximum(this, target, "melee", weapon)
             if (Hit.success(this, target, "melee", weapon, special = true)) {
-                hit1 = random.nextInt(maxHit / 2, maxHit - 10)
+                hit1 = random.nextInt(maxHit / 2, maxHit)
                 hit2 = hit1 / 2
                 hit3 = hit2 / 2
-                hit4 = hit3 + if (random.nextBoolean()) 10 else 0
+                hit4 = hit3 + 1
```
GregHib: `+ 10` because of constitution update?

game/src/main/kotlin/content/skill/melee/weapon/special/DragonClaws.kt
```diff
@@ -14,32 +15,39 @@ class DragonClaws : Script {
...
             }
 
             hit(target, damage = hit1)
             hit(target, damage = hit2)
-            hit(target, damage = hit3, delay = 30)
-            hit(target, damage = hit4, delay = 30)
+            target.strongQueue("claws_second_pair", 0) {
+                source.hit(target, weapon = weapon, damage = hit3)
+                source.hit(target, weapon = weapon, damage = hit4)
+            }
```
GregHib: These are the same, could be switched for `delay = TICKS.toClientTicks(1)` for clarity

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/interact/Interact.kt
```diff
@@ -86,6 +90,18 @@ open class Interact(
...
+        if (npc != null && !npc.def["allowed_under", false] && target is Character &&
+            Overlap.isUnder(npc.tile, npc.size, target.tile, target.size)) {
+            clearSteps()
+            val direction = Direction.cardinal.random(random)
+            if (canStep(direction.delta.x, direction.delta.y)) {
+                character.steps.queueStep(npc.tile.add(direction))
+            }
+            super.tick()
+            return
+        }
```
GregHib: Movement.stepOut could be reused here

GregHib (changes requested):

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/interact/Interact.kt
HarleyGilpin: Great catch.

game/src/main/kotlin/content/skill/melee/weapon/special/DragonClaws.kt
HarleyGilpin: After closer look this is incorrect. I've rewritten the spec formula.

HarleyGilpin:
@GregHib - I have addressed your code review notes. I also added a bug fix that caused game crashes when collapsing tabs and fixed bank tab creation that wasn't in the original PR: https://github.com/GregHib/void/pull/949/commits/78ff78bcf4451bcdba7f89c3054a05f112bf2b27

GregHib (approved):

PR #952, opened 2026-04-16
Barbarian village cleanup

M1krofin:
Been a while since last pull request. This was supposed to be a quick cleanup but those sound id's take a long time to find.

Added Unicorn combat with sounds
Added Unicorn spawn (jagex removed one in EOC don't know why)
Added Barbarian combat with sounds and warcry
Added item spawns
Added Barbarian village npc dialogue
Added Barbarian village examines which adds barbarian anvil
Fixed pottery item on object


jarryd229:
you mite get lucky and find some sound ids here but they mite not be right https://oldschool.runescape.wiki/w/List_of_sound_IDs

game/src/main/kotlin/content/skill/crafting/Pottery.kt
```diff
@@ -25,7 +25,7 @@ class Pottery : Script {
         }
 
         itemOnObjectOperate(obj = "pottery_oven*", arrive = false) { (target, item) ->
-            if (!item.def.contains("pottery")) {
+            if (!item.id.contains("unfired")) {
```
GregHib:
```suggestion
Rows.getOrNull("pottery.${item.id}") ?: return@itemOnObjectOperate
```

As there's some which don't contain "unfired" like `pot_lid`

M1krofin: Good point

GregHib (approved):

PR #957, opened 2026-04-22
Fix/freeze blocks movement

Tyluur:
## Summary
Players could still walk to trees, NPCs, and other interactables while under the freeze effect. Walk packets were blocked, but object/NPC interactions queued a new path on the next tick because the engine `Movement` never checked the freeze timer.

This PR makes freeze authoritative in the engine and blocks it at the input layer.

## Changes
- **content/entity/Movement.kt**
    - Reject `instruction<Walk>` when `hasClock("movement_delay")` is true
    - Show "A magical force stops you from moving." and return early

- **engine/.../mode/move/Movement.kt**
    - Add freeze guard at top of `tick()`
    - `if (character.hasClock("movement_delay")) { character.steps.clear(); return }`
    - Removes previous var-based check (`get("movement_delay", -5)`) which never fired because freeze uses a clock, not a var
    - Always clears steps (no `noCollision` bypass) – matches 2009-2011 behaviour

## Why hasClock?
`Freeze` starts `movement_delay` as a clock, not a variable. The engine cannot depend on the game-module `player.frozen` extension, so we check the clock directly. This keeps engine ↔ game separation clean.

## Testing
1. `::freeze 20` on self
2. Click ground → blocked with message
3. Click tree / bank booth / NPC → no movement, no path queued
4. Start walking to object, cast bind mid-path → stops on current tile next tick
5. After freeze expires → movement resumes normally

No regressions to forced movement (cutscenes/teleports use `noCollision` steps but are also cleared – intentional, freeze should override player-controlled forced walks).

Closes the last known freeze bypass.

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/client/instruction/handle/InterfaceOnPlayerOptionHandler.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/client/instruction/handle/InterfaceOnPlayerOptionHandler.kt
@@ -1,11 +1,14 @@
 package world.gregs.voidps.engine.client.instruction.handle
 
+import com.github.michaelbull.logging.InlineLogger
 import world.gregs.voidps.engine.client.instruction.InstructionHandler
 import world.gregs.voidps.engine.client.instruction.InterfaceHandler
 import world.gregs.voidps.engine.client.ui.closeInterfaces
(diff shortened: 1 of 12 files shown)


GregHib:
Also closing for #959 the issue was more that `movement_delay` was being ignored not that there wasn't checks for it

PR #967, opened 2026-04-30
Add PvP bots for Clan Wars Free-For-All

HarleyGilpin:
## Summary

 Adds bots that fight each other in the Clan Wars Free-For-All arenas
 (Safe and Dangerous). Bots use prayers, switch styles, drink potions,
 eat food, cast specs, and retreat when low. Dangerous arena follows
 wilderness PvP rules.

 ### What's included

 - Bots for both FFA arenas, with configurable population per arena
 - Realistic combat behaviour: prayers, style/spec swaps, vengeance,
   potion drinking, food eating, anti-clump movement
 - Dangerous arena follows wilderness PvP rules, players are skulled,
   kits drop, no gravestone
 - Drops from a bot kill are private to the killer, so bots don't
   become loot piñatas
 - Performance work: zone-based combat scans, deferred lookups,
   recorded-tile loot pickup, and a `/bot_stress` admin command for
   benchmarking
 - 2,400+ lines of new tests

 ### Configuration (`game/src/main/resources/game.properties`)

 - `bots.pvp.clan_wars_ffa_safe.count`, bots in Safe arena
 - `bots.pvp.clan_wars_ffa_dangerous.count`, bots in Dangerous arena
 - `bots.pvp.clan_wars_ffa_*.spawnSeconds`, spawn cadence

 Set either count to `0` to disable that arena's bots.

 ## Test plan

 - [X] `./gradlew :game:test :engine:test` passes
 - [X] `./gradlew spotlessCheck` passes
 - [X] **Safe arena**: enter Safe portal, confirm bots are present and
       fighting. Die and confirm no items drop on death.
 - [X] **Dangerous arena**: enter Dangerous portal, confirm wilderness
       rules apply, kit drops on death, no gravestone.
 - [X] **Loot privacy**: have two bots kill each other while standing
       nearby; loot from the loser should not visible.
 - [X] **Retreat**: drain a bot's food, it should retreat using a games necklace teleport.
 - [X] **Hybrid switching**: a melee+magic loadout bot should cast
       vengeance off cooldown, switch to magic when kited.
 - [X] **Performance**: run `/bot_stress` with 1,000 bots and confirmed BotMetrics performance is appropriate.

 ## Notes for reviewer

I wanted to flag upfront that I have mixed feelings about the approach and would value your thoughts.

The longer-term goal is to extend this to open wilderness bots and to populate other minigames like PC. Right now, the behavior for each combat bot is driven by configs in data/bot/pvp.combat.templates.toml, which has gotten a bit complex. I suspect there's a cleaner way to architect this, so I'm genuinely open to any criticism or suggestions you have.

No rush on the review. I'd rather take the time to get it right than ship something we'll want to redo later.


data/bot/clan_wars_tiers.tables.toml
```diff
@@ -0,0 +1,65 @@
+[clan_wars_tiers]
+row_id = "string"
+combat_style = "string"
+skills = "list<string>"
```
GregHib, line 4:
```suggestion
skills = "list<skill>"
```

game/src/main/kotlin/content/bot/combat/ClanWarsBotContext.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    private fun RowDefinition.toCombatTier(): CombatTier {
+        val skillNames = stringList("skills")
+        val values = intList("levels")
+        require(skillNames.size == values.size) { "clan_wars_tiers.$rowId: skills/levels size mismatch." }
+        val levels = LinkedHashMap<Skill, Int>(skillNames.size)
+        for ((index, name) in skillNames.withIndex()) {
+            val skillId = Skill.map[name] ?: error("clan_wars_tiers.$rowId: unknown skill '$name'.")
+            levels[Skill.all[skillId]] = values[index]
+        }
```
GregHib, line 100:
```suggestion
        val skills = skillList("skills")
        val values = intList("levels")
        require(skills.size == values.size) { "clan_wars_tiers.$rowId: skills/levels size mismatch." }
        val levels = LinkedHashMap<Skill, Int>(skills.size)
        for ((index, skill) in skills.withIndex()) {
            levels[skill] = values[index]
        }
```

data/minigame/clan_wars/clan_wars.setups.toml
```diff
@@ -0,0 +1,68 @@
...
+actions = [
+    { interface = { option = "Clan Wars", id = "worn_equipment:amulet_slot:games_necklace_8" } },
+    { wait = { ticks = 5 } },
+]
+produces = [
+    { area = "clan_wars_teleport" },
+]
+
+[enter_clan_wars_ffa_safe]
+weight = -10
```
GregHib, line 15: Don't think the weight needs to be negative, 0 would be fine?

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/player/Players.kt
```diff
@@ -71,6 +71,34 @@ object Players : Iterable<Player>, CharacterSearch<Player> {
...
 
+    /**
+     * Non-allocating chebyshev-radius scan centred on [center]. Iterates only the zones whose
+     * bounds overlap `[center.x ± radius]` × `[center.y ± radius]` on [center]'s level, then
+     * filters each zone's players by exact tile bounds.
+     *
+     * Dramatically cheaper than calling [at] per-tile in a spiral: one zone lookup per overlapping
+     * zone (≤ 25 for radius 15) versus 961 tile lookups each scanning the full zone.
+     */
+    inline fun forEachInRadius(center: Tile, radius: Int, action: (Player) -> Unit) {
```
GregHib: This doesn't need to be inline? and indexArray/map can stay private

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/player/Players.kt
```diff
@@ -71,6 +71,34 @@ object Players : Iterable<Player>, CharacterSearch<Player> {
...
+                map.onEach(Zone.id(zx, zy, level)) { index ->
+                    val player = indexArray[index] ?: return@onEach
+                    val dx = player.tile.x - center.x
+                    if (dx < -radius || dx > radius) return@onEach
+                    val dy = player.tile.y - center.y
+                    if (dy < -radius || dy > radius) return@onEach
+                    action(player)
+                }
+            }
+        }
```
GregHib, line 101: (Double check this behaviour)
```suggestion
        val level = center.level
        val area = center.toCuboid(radius)
        for (zone in center.toCuboid(radius).toZones(level)) {
                map.onEach(zone) { index ->
                    val player = indexArray[index] ?: return@onEach
                    if (player.tile !in area) {
                      return@onEach
                    }
                    action(player)
                }
        }
```

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/CharacterIndexMap.kt
```diff
@@ -51,8 +52,11 @@ class CharacterIndexMap(size: Int) {
         current.fill(INVALID)
     }
 
-    fun onEach(id: Int, action: (Int) -> Unit) {
-        table.get(id)?.onEach(action)
+    inline fun onEach(id: Int, action: (Int) -> Unit) {
```
GregHib: Again with the inline and public, keep these things private

game/src/main/kotlin/content/bot/behaviour/action/ActionParser.kt
```diff
@@ -225,6 +373,7 @@ sealed class ActionParser {
 
         private val parsers = mapOf(
             "npc" to InteractNpcParser,
+            "player" to InteractPlayerParser,
```
GregHib, line 370: player_att or player_combat ? This parser wouldn't be very relevant for trading or following a player/general interaction

game/src/main/kotlin/content/bot/behaviour/action/ActionParser.kt
```diff
@@ -42,15 +45,159 @@ sealed class ActionParser {
...
+        override val optional = setOf("if", "success")
+
+        override fun parse(map: Map<String, Any>): BotAction {
+            val item = map["item"] as String
+            val area = map["area"] as String
+            val condition = requirement(map, "if").singleOrNull()
+            val success = requirement(map, "success").singleOrNull()
+            return BotJewelleryTeleport(item, area, condition, success)
         }
     }
```
GregHib, line 196: Can maybe be made more generic, does it need to specify the method, surely using the usual navigation system can figure out that the jewellery is the best route to escape?

game/src/main/kotlin/content/bot/behaviour/action/BotCastSpell.kt
```diff
@@ -0,0 +1,220 @@
...
+    val kite: Boolean = true,
+    val area: String? = null,
+) : BotAction {
+    override fun update(bot: Bot, world: BotWorld, frame: BehaviourFrame): BehaviourState? {
+        // Success first so a retreat-by-teleport (bot now outside `area`) can complete the
+        // activity even at low HP — otherwise eat() spins forever when food is exhausted.
+        if (success?.check(bot.player) == true) return BehaviourState.Success
+        if (healPercentage > 0 && bot.levels.get(Skill.Constitution) <= bot.levels.getMax(Skill.Constitution) * healPercentage / 100) {
+            return eat(bot, world)
+        }
```
GregHib: I'm thinking healing/eating should be a reactive action as it's quite generic and applies to all combat scenarios, would also mean removing the `healPercentage` from all these actions which would tidy them up a bit

game/src/main/kotlin/content/bot/behaviour/action/BotCastSpell.kt
```diff
@@ -0,0 +1,220 @@
...
+
+    private fun maybeKite(bot: Bot, world: BotWorld, target: Player) {
+        if (!kite || !target.frozen) return
+        if (bot.tile.distanceTo(target.tile) > 2) return
+        val dx = (bot.tile.x - target.tile.x).coerceIn(-1, 1)
+        val dy = (bot.tile.y - target.tile.y).coerceIn(-1, 1)
+        if (dx == 0 && dy == 0) return
+        val kx = bot.tile.x + dx * 2
+        val ky = bot.tile.y + dy * 2
+        val dest = Tile(kx, ky, bot.tile.level)
```
GregHib, line 88: (Double check the logic)
```suggestion
        val dir = bot.tile.delta(target.tile).toDirection()
        if (dir == Direction.None) {
            return 
        }
        val dest = bot.tile.add(dir).add(dir)
```

game/src/main/kotlin/content/bot/behaviour/action/BotLootStrategy.kt
```diff
@@ -0,0 +1,50 @@
...
+    fun accepts(item: FloorItem): Boolean = when (this) {
+        SURVIVAL -> isConsumable(item)
+        DEFAULT, WEALTH -> true
+    }
+
+    /** Whether candidates should be gathered and ranked instead of taking the first match. */
+    fun ranks(): Boolean = this == WEALTH
+
+    /** Total coin value to compare candidates by; higher is better. */
+    fun score(item: FloorItem): Long = item.value
```
GregHib, line 31: I wonder if it's worth making looting it's own activity. On one hand it would keep the code duplication down and remove it from the combat actions, on the other hand having it switch from fighting to looting after the target is killed would prevent survival looting of nearby (though this could also be a separate reactive action?) wdyt?

game/src/main/kotlin/content/bot/behaviour/condition/BotTargetArmorType.kt
```diff
@@ -0,0 +1,26 @@
+package content.bot.behaviour.condition
+
+import content.bot.bot
+import content.bot.isBot
+import world.gregs.voidps.engine.entity.character.mode.interact.PlayerOnPlayerInteract
+import world.gregs.voidps.engine.entity.character.player.Player
+import world.gregs.voidps.engine.entity.character.player.equip.equipped
+import world.gregs.voidps.network.login.protocol.visual.update.player.EquipSlot
+
+data class BotTargetArmorType(val equals: Set<String>) : Condition(1) {
```
GregHib: For consistency 😜 
```suggestion
data class BotTargetArmourType(val equals: Set<String>) : Condition(1) {
```

game/src/main/kotlin/content/bot/behaviour/perception/BotClanWarRole.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        private fun healingItems(player: Player): Int {
+            val inv = player.inventory
+            var count = 0
+            for (index in inv.indices) {
+                val item = inv[index]
+                val id = item.def.stringId
+                if (id == "shark" || id.startsWith("saradomin_brew_")) {
+                    count += item.amount
+                }
```
GregHib: Maybe a data table with different options? Might be a good idea to reuse the logic when finding the index of a healing item to consume as well (and support drinking to heal not just "Eat")

game/src/main/kotlin/content/bot/behaviour/action/ActionParser.kt
```diff
@@ -42,15 +45,159 @@ sealed class ActionParser {
...
+        override val optional = setOf("delay", "success", "radius", "heal_percent", "target_score", "family", "kite", "area")
+
+        @Suppress("UNCHECKED_CAST")
+        override fun parse(map: Map<String, Any>): BotAction {
+            val delay = map["delay"] as? Int ?: 0
+            val success = requirement(map, "success").singleOrNull()
+            val radius = map["radius"] as? Int ?: 10
+            val healPercent = map["heal_percent"] as? Int ?: 40
+            val rawScore = map["target_score"] as? List<Map<String, Any>>
+            val scorer = rawScore?.let { UtilityCurveParser.parseScorer(it) }
```
GregHib: Target score and the utility system does not seem to be used at all??

GregHib (commented):
First pass.

> I wanted to flag upfront that I have mixed feelings about the approach and would value your thoughts.

> The longer-term goal is to extend this to open wilderness bots and to populate other minigames like PC. Right now, the behavior for each combat bot is driven by configs in data/bot/pvp.combat.templates.toml, which has gotten a bit complex. I suspect there's a cleaner way to architect this, so I'm genuinely open to any criticism or suggestions you have.


I think minigame_combat.templates.toml just needs splitting up, maybe into pvp styles?

I think for now it's okay and these larger complexity issues don't need addressing in this PR as some of them are quite fundamental to the bot system as a whole.

There's a few things the bot system could do with which this work highlights:
1. Reusable action templates could tidy it up i.e. not having to write 
`{ pray = { id = "deflect_melee", if = { attacker_style = { equals = "melee" } } } },` in full everytime.
2. A proper target finding system (there's probably some cool things we could do with influence mapping)
3. A gear/loadout best-fit kind of system or extension on the current (like what you've started with any but for setup)

game/src/main/resources/game.properties
```diff
@@ -264,10 +264,10 @@ events.tearsOfGuthix.active=true
 #===================================
 
 # The number of AI-controlled bots spawned on startup
-bots.count=30
+bots.count=20
 
 # Frequency between spawning bots on startup
-bots.spawnSeconds=60
+bots.spawnSeconds=1
```
GregHib:
```suggestion
bots.spawnSeconds=60
```

Good to keep this high by default to avoid causing spikes on startup

game/src/main/resources/game.properties
```diff
@@ -296,6 +296,18 @@ bots.shortcuts=shortcuts.toml
...
+bots.combat.clan_wars_ffa_safe.count=9
+
+# Seconds between combat bot spawns for the FFA Safe arena while ramping to count.
+bots.combat.clan_wars_ffa_safe.spawnSeconds=1
+
+# Target population of combat bots inside the Clan Wars FFA Dangerous arena. 0 disables.
+bots.combat.clan_wars_ffa_dangerous.count=9
+
+# Seconds between combat bot spawns for the FFA Dangerous arena while ramping to count.
+bots.combat.clan_wars_ffa_dangerous.spawnSeconds=1
```
GregHib:
```suggestion
bots.combat.clan_wars_ffa_safe.spawnSeconds=60

# Target population of combat bots inside the Clan Wars FFA Dangerous arena. 0 disables.
bots.combat.clan_wars_ffa_dangerous.count=9

# Seconds between combat bot spawns for the FFA Dangerous arena while ramping to count.
bots.combat.clan_wars_ffa_dangerous.spawnSeconds=60
```

game/src/main/resources/game.properties
HarleyGilpin: Increased back to 1 minute, I also disabled CW FFA bot spawns by default.

game/src/main/resources/game.properties
HarleyGilpin: I changed count for both to 0 to disable this feature as default.

game/src/main/kotlin/content/bot/behaviour/action/ActionParser.kt
HarleyGilpin: The scorer is used. It's live in `BotCastSpell.handleCombat` and `BotFightPlayer.handleEngaged` / `pickTarget`

HarleyGilpin:
Thanks for the review! Pushed a follow-up that addresses most of it.

### Applied

- `clan_wars_tiers.tables.toml`: `skills = "list<skill>"` now, and `ClanWarsBotContext.toCombatTier` goes through a new `Tables.skillList` / `RowDefinition.skillList` helper.
- `clan_wars.setups.toml` weights are positive again: `enter_clan_wars_*` at 10, `teleport_clan_wars_games_necklace` at 20. I also moved `DynamicResolvers.go_to_area` from `-1` to `100` so explicit setups always beat the generic pathfind fallback (the original negative was load-bearing because of that).
- `BotTargetArmorType` is now `BotTargetArmourType` (file, class, and TOML key).
- `bots.spawnSeconds` and per-arena `bots.combat.*.spawnSeconds` are back to 60.
- `Players.forEachInRadius`: dropped `inline`, set `indexArray` / `map` back to `private`, and switched to `center.toCuboid(radius).toZones(...)` with `tile !in area`. `CharacterIndexMap.onEach` is also non-inline now.
- `"player"` action key renamed to `"player_combat"`. The parser still hard-requires `option = "Attack"`.
- `BotCastSpell.maybeKite` now uses `tile.delta(target).toDirection()` followed by `tile.add(direction).add(direction)`.
- **Reactive eat extracted.** New `BotReactiveEat` plus `data/bot/heal_items.tables.toml` (data-driven food/drink list). `healPercentage` is gone from `BotFightPlayer`, `BotCastSpell`, `BotFightNpc`, and their parsers. Every combat template now leads its `reactive = [...]` with `{ eat = { if = { skill_percent = { id = "constitution", max_percent = N } } } }`. The reactive gates on `food_delay` / `drink_delay` / `combo_delay` so it doesn't keep extending `action_delay` mid-consume.
- `BotClanWarRole.healingItems` reads from that same heal-items table now (no more hard-coded `shark` / `saradomin_brew_*`).
- Split `minigame_combat.templates.toml` into `pvp_safe.templates.toml` and `pvp_dangerous.templates.toml`.
- **Survival-loot reactive.** New `BotReactiveLoot` (`loot` parser key), wired into the four dangerous templates. Skips when the bot is frozen, when the consume window is active, or when inventory is full, and respects ownership.

### Re: your `target_score` question

The scorer is used: it's live in `BotCastSpell.handleCombat` and in `BotFightPlayer.handleEngaged` / `pickTarget`. 

### Deferred to follow-up PRs

- Jewellery-teleport into a generic nav-driven retreat. Greg wants the nav graph to model "any teleport that produces area X" first.
- Reusable reactive blocks (the `{ pray = { id = "deflect_melee", ... } }` duplication you flagged). Needs a parser-layer feature, so I'd rather do it as a focused PR.

GregHib:
There's a few comments that were missed (hidden in githubs see-more button).

GregHib (approved):
Just finished up the last bits to get it over the line before it starts conflicting too much with other changes

PR #983, opened 2026-05-15
Feat: Add Summoning Pets

HarleyGilpin:
# Add pets system
 
## Summary
 
Adds the pets system to Void. Pets are non-combat follower NPCs that follow the player, grow over time, get hungry, can be fed, and can be talked to. A player can have one follower at a time (pet or combat familiar, not both).

image

## Features
 
- **85 pet variants** registered through a single TOML registry: kittens, cats, hellcat, six dog breeds (three colours each), four baby dragons, penguin, raven, ten monkey colours, squirrel, vulture, chameleon, giant crab, gecko, platypus, raccoon, godbirds, broav, parrot, clockwork cat, and phoenix egglings.
- **Spawning & interaction**: Dropping a pet item spawns it next to the player and starts it following. Right-click exposes Pick-up (returns the item to inventory) and Talk-to (plays dialogue). Ownership is enforced. Drop, pick-up, and feed play the climb-down animation.
- **Hunger & growth**: Tick every 30 seconds per pet. Feeding the correct food lowers hunger by 15. Max hunger causes the pet to run away; max growth triggers metamorphosis to the next life stage. Phoenix egglings, clockwork cat, broav, and ex-ex-parrot intentionally skip ticking.
- **Chat**: Occasional overhead idle chatter, hungry-threshold phrases, and Talk-to dialogue, all configured per pet. Lines prefixed `pet:` render as overhead speech; the rest as chatbox statements.
- **Summoning orb**: Activates for pets the same way as familiars. Left-click options (Follower Details, Call Follower, Dismiss) route to whichever slot is occupied. Follower Details shows name, chathead, and packed growth/hunger bars.
- **Persistence**: Active pet item and a compact per-pet stats blob persist on the player. On login the pet respawns at the player's tile and resumes ticking.
- **Incubators**: Taverley and Yanille incubators accept eggs, transform visually while incubating, and hatch into the matching baby pet after the configured time. Incubator state persists across logout.

## Notable files
 
- `game/src/main/kotlin/content/skill/summoning/pet/`: new package containing the registry, lifecycle, timers, feeding, incubator, dialogue, and persistence helpers.
- `game/src/main/kotlin/content/skill/summoning/Summoning.kt`: pet/familiar mutual exclusion and orb-option routing.
- `game/src/main/kotlin/content/area/misthalin/varrock/Gertrude.kt`: Minimal dialogue for Gertrude.
- `data/skill/summoning/pet/`: pet registry, incubator egg registry, NPC and item string-id overrides, incubator object string-ids, pet/incubator variable definitions.
- `data/skill/summoning/summoning.ifaces.toml`, `data/entity/player/modal/toplevel/gameframe.ifaces.toml`, `data/skill/summoning/summoning.varps.toml`: missing interface option and varp declarations.
- `engine/src/main/kotlin/world/gregs/voidps/engine/data/PlayerSave.kt`: bumps the save reader buffer for the pet stats blob.
- `game/src/test/kotlin/content/skill/summoning/pet/`: regression tests covering drop-to-summon, logout persistence, mutual exclusion, ownership, overhead chat, Talk-to dialogue, incubator flow, and orb/Follower Details routing.
## Test plan
 
- [x] Drop a pet item; confirm it spawns and follows.
- [x] Right-click Pick-up; confirm the item returns to inventory.
- [x] Right-click Talk-to; confirm dialogue plays for several pet types.
- [x] Second player tries to Pick-up your pet; confirm refusal.
- [x] Try to summon a familiar while a pet is active (and vice versa); confirm only one follower at a time.
- [x] Feed a pet its correct food; confirm hunger drops and the climb-down animation plays.
- [x] Wait for hunger and growth ticks; confirm hungry-phrase overhead chat, idle chatter, run-away at max hunger, and metamorphosis at max growth.
- [x] Open the summoning orb with a pet active; confirm Follower Details, Call Follower, and Dismiss all work.
- [x] Log out with an active pet and an egg incubating; log back in; confirm both resume correctly.
- [x] Use each egg type on Taverley and Yanille incubators; confirm hatching into the matching baby pet.


game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
```diff
@@ -0,0 +1,147 @@
...
+private fun endKey(suffix: String) = "incubator_end_$suffix"
+
+private const val EMPTY = 0
+private const val INCUBATING = 1
+private const val FINISHED = 2
+
+class Incubator(private val definitions: IncubatorDefinitions) : Script {
+
+    init {
+        objectOperate("Inspect", "incubator_*") {
```
GregHib, line 30: If the names of the incubator are _taverley/_yanille you can use it.target.id.removePrefix("incubator_") to get the suffix and construct the name of the varbit, removing regionVarbits entirely.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
```diff
@@ -0,0 +1,147 @@
...
+                active = true
+            }
+            if (!active) Timer.CANCEL else Timer.CONTINUE
+        }
+
+        playerSpawn {
+            for (region in regionVarbits.values) {
+                if (get(eggKey(region.suffix), "").isNotBlank()) {
+                    timers.start("incubator_check")
+                    return@playerSpawn
```
GregHib:
```suggestion
                if (contains(eggKey(region.suffix))) {
                    timers.restart("incubator_check")
```

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
```diff
@@ -0,0 +1,147 @@
...
+                }
+                // Ensure stale varbit doesn't show occupied incubator after a wipe.
+                val state = get<Int>(region.varbit) ?: EMPTY
+                if (state != EMPTY) set(region.varbit, EMPTY)
```
GregHib: That can never be? Just make sure to sendVariable for each varbit instead?

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+    }
+
+    private fun Player.currentState(region: IncubatorRegion): Int {
+        val eggId = get(eggKey(region.suffix), "")
+        if (eggId.isBlank()) return EMPTY
+        val state = get<Int>(region.varbit) ?: INCUBATING
+        if (state == FINISHED) return FINISHED
+        if (System.currentTimeMillis() >= get(endKey(region.suffix), 0L)) return FINISHED
```
GregHib: Making the varbits `format = "list"` would simplify the egg state a lot.

Also better to use clocks for the end with `remaining()` rather than tracking current millis

game/src/main/kotlin/content/skill/summoning/pet/IncubatorDefinitions.kt
GregHib, line 1: Can be removed entirely, and the Tables and Rows used directly (see PollnivneachDungeon.kt for example)

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,89 @@
...
+    private suspend fun Player.openMenu(kitten: NPC) {
+        choice("Interact with Kitten") {
+            option("Stroke") { stroke(kitten) }
+            option("Chase vermin") { chaseVermin(kitten) }
+            option("Shoo away") { shooConfirm(kitten) }
+        }
+    }
+
+    private suspend fun Player.stroke(kitten: NPC) {
+        kitten.watch(this)
```
GregHib: Probably needs a kitten.steps.clear() to avoid it moving when you stroke (like in your gif)

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
```diff
@@ -0,0 +1,269 @@
...
+    val itemIntId = ItemDefinitions.getOrNull(itemStringId)?.id ?: 0
+    set("follower_details_name", itemIntId)
+    set("follower_details_chathead", pet.def.id)
+    set("follower_details_chathead_animation", pet.id)
+    // On relogin the persisted values match what we just wrote, so
+    // `PlayerVariables.set` short-circuits and never transmits a packet.
+    // Force-send so the client always receives the current state.
+    variables.send("follower_details_name")
+    variables.send("follower_details_chathead")
+    variables.send("follower_details_chathead_animation")
```
GregHib: These are only necessary in a playerSpawn {}

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
```diff
@@ -0,0 +1,269 @@
...
+        npcOperate("Talk-to", npcId) { interact ->
+            val owner = pet
+            if (owner == null || owner.index != interact.target.index) {
+                message("This isn't your pet.")
+                return@npcOperate
+            }
+            talkToPet(def, owner)
+        }
+    }
+}
```
GregHib: It's better have a single `npcOperate("Pick-up", "npc1,npc2,npc3")` for each type of operation and lookup the definition inside the method, rather than loop over every one; as the former comes with more memory + startup overhead.

game/src/main/kotlin/content/skill/summoning/pet/PetDefinitions.kt
GregHib, line 1: Can also almost be entirely removed, if the baby/gown/overgrown item and npc ids are named predictably you wouldn't need to look them up, it'd just be a case of doing `npc.id.replace("_baby", "_grown")` etc...

game/src/main/kotlin/content/skill/summoning/pet/PetState.kt
```diff
@@ -0,0 +1,57 @@
...
+private fun Player.writePetStats(map: Map<String, PetStats>) {
+    val encoded = map.entries
+        .filter { !it.value.isDefault() }
+        .joinToString(",") { (id, s) -> "$id:${s.hunger}:${s.growth}:${s.warn}" }
+    set("pet_stats", encoded)
+}
+
+fun Player.getPetHunger(id: String): Double = readPetStats()[id]?.hunger ?: 0.0
+fun Player.getPetGrowth(id: String): Double = readPetStats()[id]?.growth ?: 0.0
+fun Player.getPetWarn(id: String): Int = readPetStats()[id]?.warn ?: 0
```
GregHib: Just use a predictable key, i.e. `"pet_${id}_hunger", "pet_${id}_growth", "pet_${id}_warn"`. All this encoding/decoding is unnecessary complexity

game/src/main/kotlin/content/skill/summoning/Summoning.kt
```diff
@@ -222,14 +238,21 @@ class Summoning : Script {
...
-                        option("No.")
+                    } else if (pet != null) {
+                        pickupPet()
                     }
                 }
-                "Dismiss Now" -> dismissFamiliar()
+                "Dismiss Now" -> when {
+                    follower != null -> dismissFamiliar()
+                    pet != null -> pickupPet()
+                }
```
GregHib, line 280: Pets have a different interface from familiars, id 663

GregHib (changes requested):

HarleyGilpin:
Review feedback addressed in 1b3959f8. Per-comment notes:

1. **regionVarbits removal**: collapsed to a one-line `tile.region.id -> suffix` helper; cache transforms (28336/28359) are shared between sites so right-click `target.id` can't carry the region by itself.
2. **`contains` + `restart`**: applied verbatim.
3. **Dead varbit-reset branch**: dropped; `playerSpawn` now just `variables.send`s the state varbits.
4. **list-format varbit + clocks**: `incubator_state_*` is `format = "list"` (`empty/incubating/finished`); `incubator_end_*` is a tick clock via `start`/`remaining`. Magic ints + `System.currentTimeMillis()` are gone.
5. **`IncubatorDefinitions` deletion**: gone; reads via `Tables.get("incubator_eggs").rows()` / `Rows.getOrNull("incubator_eggs.$id")`.
6. **Kitten stops during Stroke**: `steps.clear()` + `mode = EmptyMode` for the animation, restored to `Follow` afterwards.
7. **`variables.send` only on `playerSpawn`**: moved out of `updatePetInterface`.
8. **Comma-csv ops**: one `npcOperate` per option, one `itemOption` per option; same for `PetFeeding`.
9. **`PetDefinitions` thinned**: now a Koin singleton with a `by lazy` row list and three linear-search lookups. Full removal via naming convention needs a data rename pass first (`pet_kitten`/`pet_cat`/`overgrown_cat` are inconsistent, plus hellcat/clockwork-cat/eggling); happy to follow up.
10. **`pet_stats` blob replaced**: per-pet `pet_${id}_hunger`/`_growth`/`_warn` vars declared in a generated `pet_state.vars.toml`; added engine-side `DoubleValues` so 0.0 defaults prune like int/bool.
11. **Interface 663**: `updatePetInterface` opens `pet_details`; the dismiss handler in `Summoning.kt` now globs `*_details:dismiss` so the pet path inherits it.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
HarleyGilpin: The cache transforms (28336/28359) are shared between Taverley and Yanille spawns so right-click `target.id` resolves to `incubator_idle`/`incubator_active` regardless of region — can't derive the suffix from that alone. Collapsed `regionVarbits` + `IncubatorRegion` down to a one-line `tile.region.id -> suffix` helper and wired all the varbit/var lookups through `incubator_X_$suffix`.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
HarleyGilpin: Applied in 1b3959f8 `contains` + `timers.restart`.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
HarleyGilpin: `playerSpawn` now just `variables.send`s each state varbit and starts the check timer if an egg is persisted.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
HarleyGilpin: `incubator_state_*` is now `format = "list"` with `empty/incubating/finished` and `incubator_end_*` is a tick clock via `start` / `remaining` — the `EMPTY/INCUBATING/FINISHED` int constants and the `System.currentTimeMillis()` branch are gone.

game/src/main/kotlin/content/skill/summoning/pet/IncubatorDefinitions.kt
HarleyGilpin: Incubator.kt` reads `Tables.get("incubator_eggs").rows()` and `Rows.getOrNull("incubator_eggs.$id")` directly, matching the `PollnivneachDungeon` pattern.

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: steps.clear()` plus switching the kitten to `EmptyMode` for the stroke animation, then restoring `Follow` afterwards.

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
HarleyGilpin: the three `variables.send` calls now live in `playerSpawn` only; the live-summon path relies on normal `set` packet emission.

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
HarleyGilpin: Collapsed in 1b3959f8. Two `npcOperate`s (`Pick-up`, `Talk-to`) and two `itemOption`s (`Drop`, `Release`) now key off a single comma-separated id list with the definition resolved inside the handler. Same for `PetFeeding.itemOnNPCOperate`.

game/src/main/kotlin/content/skill/summoning/pet/PetDefinitions.kt
HarleyGilpin: Thinned in 1b3959f8. `PetDefinitions` is now a Koin singleton with a single `by lazy { Tables.get("pets").rows().map(PetDefinition::from) }` list and three linear-search lookups — no `Object2ObjectOpenHashMap` caches, no `load(path)` step. Full removal via naming convention will need a data rename pass first (`pet_kitten -> pet_cat`, `overgrown_cat` vs `pet_cat_overgrown`, the hellcat/clockwork-cat/eggling odd ones); happy to open a follow-up for that.

game/src/main/kotlin/content/skill/summoning/pet/PetState.kt
HarleyGilpin: Done in 1b3959f8. `pet_stats` is gone; per-pet `pet_${id}_hunger` / `_growth` / `_warn` vars are declared in a generated `pet_state.vars.toml` (255 entries from `pets.tables.toml`). Added `DoubleValues` to the engine so 0.0 defaults prune like int/bool — no more encode/decode in `PetState.kt`.

game/src/main/kotlin/content/skill/summoning/Summoning.kt
HarleyGilpin: Routed in 1b3959f8. `updatePetInterface` now opens `pet_details` (663), `deactivateSummoningOrb` closes both ifaces, and the `familiar_details:dismiss` handler in `Summoning.kt` now globs `*_details:dismiss` so the pet path picks it up.

data/entity/npc/monster/draconic/dragon.drops.toml
```diff
@@ -146,12 +147,19 @@ drops = [
   { id = "blue_charm", amount = 3, chance = 13 },
 ]
 
+[black_dragon_pet_egg]
+roll = 2000
+drops = [
+  { id = "black_dragon_egg" },
```
GregHib:
```suggestion
  { id = "black_dragon_egg", skill = "summoning", equals = 99 },
```

Added support in #984

data/skill/summoning/pet/pet.npcs.toml
```diff
@@ -0,0 +1,700 @@
+[pet_cat_baby]
+id = 761
+options = { "Pick-up" = 1, "Talk-to" = 2, "Interact-with" = 3 }
```
GregHib: These aren't doing anything (options are only necessary for interfaces to encode context menu unlocks)
```suggestion
```

engine/src/main/kotlin/world/gregs/voidps/engine/client/variable/VariableValues.kt
```diff
@@ -15,6 +15,7 @@ sealed class VariableValues {
         @Suppress("UNCHECKED_CAST")
         operator fun invoke(values: Any?, format: String?, default: Any?): VariableValues = when (format ?: default?.apply { this::class.java.simpleName.lowercase() }) {
             "int", "integer" -> IntValues
+            "double" -> DoubleValues
```
GregHib: Doubles are intentionally not supported as rs doesn't store or use doubles anywhere, it's always a 1 in X value (e.g. run energy being out of a 10,000 total). So growth rates and hunger levels will all need to be converted to integers.

engine/src/main/kotlin/world/gregs/voidps/engine/data/PlayerSave.kt
```diff
@@ -223,7 +225,9 @@ data class PlayerSave(
             val ignores = ObjectArrayList<String>()
             val offers = Array(6) { ExchangeOffer.EMPTY }
             val history = ObjectArrayList<ExchangeHistory>()
-            Config.fileReader(file) {
+            // 8192 buffer accommodates long player-variable values (e.g. encoded
+            // pet_stats blob holding hunger/growth for every pet ever raised).
+            Config.fileReader(file, maxStringLength = 8192) {
```
GregHib: Please revert as no longer in use

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
```diff
@@ -0,0 +1,128 @@
...
+
+private fun eggForItem(itemId: String): RowDefinition? = Tables.get("incubator_eggs").rows().firstOrNull { it.item("egg") == itemId }
+
+class Incubator : Script {
+
+    init {
+        objectOperate("Inspect", "incubator_*") {
+            val suffix = incubatorSuffix() ?: return@objectOperate run {
+                message("This incubator isn't ready to be used.")
+            }
```
GregHib: Only `incubator_taverley` and `incubator_yanille` exist and the interaction guarantees they're in the correct location so there's no need to double check that.
```suggestion
            val suffix = it.target.id.removePrefix("incubator_")
```

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,220 @@
...
+private const val SCAN_RADIUS = 10
+private const val CHASE_RADIUS = 8
+private const val CATCH_CHANCE = 0.33
+
+private fun NPC.isRat(): Boolean {
+    val id = this.id
+    return id == "rat" ||
+        id.startsWith("rat_") ||
+        id.endsWith("_rat") ||
+        "_rat_" in id
```
GregHib: I think `return id.startsWith("rat")` should do? Otherwise it'll also match giant_rat's

game/src/main/kotlin/content/skill/summoning/pet/PetDefinitions.kt
GregHib: Yeah the renaming can be done in a follow-up, but `PetDefinitions` as a whole is still redundant as `PetDefinitions.get()` can be replaced with direct table calls like `Tables.npc("pets.${npcId}.${stage}_npc")`. PetDefinitions is just storing the same data in memory a second time.

engine/src/main/kotlin/world/gregs/voidps/engine/client/variable/VariableValues.kt
```diff
@@ -35,6 +36,11 @@ data object IntValues : VariableValues() {
     override fun toInt(value: Any) = value as Int
 }
 
+data object DoubleValues : VariableValues() {
+    override fun default() = 0.0
+    override fun toInt(value: Any) = (value as Double).toInt()
```
GregHib: This wouldn't even work, it'd always send 0 or 1. (The client has no packets for sending double variables)

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,220 @@
...
+                player<Quiz>("What do you want to do now?")
+                npc<Happy>(cat.id, "I want to go chase things, kill them and then eat them.")
+                player<Quiz>("Always with the hunting...")
+            }
+            option("That's enough talking for now.") {
+                keepGoing = false
+            }
+        }
+        if (!keepGoing) return
+    }
```
GregHib:
```suggestion
        choice("What would you like to ask?") {
            option("How are you doing?") {
                player<Quiz>("How are you doing?")
                npc<Happy>(cat.id, "I'm good. But could we go adventuring soon? I'd like to chase things and eat them.")
                talkToCatWithAmulet(cat)
            }
            option("How old are you now?") {
                player<Quiz>("How old are you now?")
                npc<Happy>(cat.id, "I'm not too old, and not too young. The perfect age really.")
                talkToCatWithAmulet(cat)
            }
            option("Where do you want to go?") {
                player<Quiz>("Where do you want to go?")
                npc<Happy>(cat.id, "Can we go to Varrock Sewer and chase some rats?")
                talkToCatWithAmulet(cat)
            }
            option("What do you want to do now?") {
                player<Quiz>("What do you want to do now?")
                npc<Happy>(cat.id, "I want to go chase things, kill them and then eat them.")
                player<Quiz>("Always with the hunting...")
                talkToCatWithAmulet(cat)
            }
            option("That's enough talking for now.")
        }
    }
```
Safer (and cleaner) than infinite loops

game/src/main/kotlin/content/skill/summoning/pet/PetState.kt
```diff
@@ -0,0 +1,29 @@
...
+fun Player.getPetWarn(id: String): Int = get("pet_${id}_warn", 0)
+
+/** Mutate the named pet's stats. Each field persists in its own player variable. */
+fun Player.updatePetStats(id: String, block: PetStats.() -> Unit) {
+    val stats = PetStats(getPetHunger(id), getPetGrowth(id), getPetWarn(id))
+    block(stats)
+    set("pet_${id}_hunger", stats.hunger)
+    set("pet_${id}_growth", stats.growth)
+    set("pet_${id}_warn", stats.warn)
+}
```
GregHib: This can all be inlined now and utilise inc/dec with max values

GregHib (changes requested):

data/entity/npc/monster/draconic/dragon.drops.toml
HarleyGilpin: Applied in 5b2c008a. The black dragon egg drop reads `skill = "summoning", equals = 99` now. Left red / blue / green un-gated since lower level summoners still need those eggs to use the incubator.

data/skill/summoning/pet/pet.npcs.toml
HarleyGilpin: Done in 5b2c008a. Every `options = { ... }` map is gone, each pet npc entry is just the id now.

engine/src/main/kotlin/world/gregs/voidps/engine/client/variable/VariableValues.kt
HarleyGilpin: Got it. Removed `DoubleValues` entirely in 5b2c008a and converted hunger / growth / growth-rate to ints on a 0..10000 scale. `growth_rate` (double) became `growth_per_tick` (int = rate * 50 * 100 per 30s tick), and the per-tick math and warn thresholds were rescaled accordingly.

engine/src/main/kotlin/world/gregs/voidps/engine/data/PlayerSave.kt
HarleyGilpin: Reverted in 5b2c008a.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
HarleyGilpin: Inlined `it.target.id.removePrefix("incubator_")` in every handler in 5b2c008a.

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: Simplified to `id.startsWith("rat")` . Drops giant_rat / warped_rat / stronghold_rat from the chase pool

game/src/main/kotlin/content/skill/summoning/pet/PetDefinitions.kt
HarleyGilpin: Deleted in 5b2c008a. New `Pets.kt` carries the `PetStage` enum plus `RowDefinition` extensions (isCatLike, stageForItem/Npc, npcFor/itemFor, nextStageItem/Npc, isFinalStage, ambientPhrases) and two lookup helpers (`petRowForItem` / `petRowForNpc`). `PetScripts`, `KittenInteract`, `PetTimers` and `PetFeeding` consume `RowDefinition` directly now and the Koin singleton is gone from `GameModules`.

engine/src/main/kotlin/world/gregs/voidps/engine/client/variable/VariableValues.kt
HarleyGilpin: Yep, gone in 5b2c008a along with the `"double"` factory mapping. Hunger and growth are ints now.

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: Applied your snippet in 5b2c008a, the `while(true)` and the `keepGoing` flag are both gone.

game/src/main/kotlin/content/skill/summoning/pet/PetState.kt
HarleyGilpin: Inlined in 5b2c008a. `PetStats` and `updatePetStats` are gone. The getters return `Int` and the callers use `inc("pet_${id}_hunger", amount, max = PET_STAT_MAX)` and `dec("pet_${id}_hunger", amount)` straight off `VariableStore`.

data/entity/npc/monster/draconic/dragon.drops.toml
GregHib: Should apply to all eggs

engine/src/main/kotlin/world/gregs/voidps/engine/data/definition/ColumnReader.kt
```diff
@@ -21,6 +21,12 @@ sealed interface ColumnReader<T : Any> {
         override fun read(reader: ConfigReader) = reader.int()
     }
 
+    object ReaderDouble : ColumnReader<Double> {
+        override val type = ColumnType.ColumnDouble
+        override fun list() = mutableListOf<Double>()
+        override fun read(reader: ConfigReader) = reader.double()
+    }
+
```
GregHib: Can be reverted/removed now

engine/src/main/kotlin/world/gregs/voidps/engine/data/PlayerSave.kt
GregHib, line 1: Just revert this whole file, the double decoding is there for backwards compatibility.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
```diff
@@ -0,0 +1,129 @@
...
+            if (!active) Timer.CANCEL else Timer.CONTINUE
+        }
+
+        playerSpawn {
+            for (suffix in SUFFIXES) {
+                if (contains("incubator_egg_$suffix")) {
+                    // GameLoop.tick resets to 0 on server boot, so the stored
+                    // tick clock is meaningless after a restart. Reset the
+                    // clock fresh for any egg the player left mid-incubation
+                    // so they keep the egg but start the countdown over.
```
GregHib: You can use `epochSeconds()` as a clock base to store them safely if that's how incubators work (I'm not sure if it is) e.g. `if (hasClock("bork_cooldown", epochSeconds())) {`

game/src/main/kotlin/content/skill/summoning/pet/PetShopOwner.kt
```diff
@@ -0,0 +1,214 @@
...
+        // interface-option packet depending on how the cache wires up each
+        // button, so register both dispatch paths against the same resume.
+        for ((breedId, index) in BREED_INDEX) {
+            continueDialogue("dialogue_pick_a_puppy:$breedId") {
+                (suspension as? Suspension.IntEntry)?.resume(index)
+            }
+            interfaceOption(id = "dialogue_pick_a_puppy:$breedId") {
+                (suspension as? Suspension.IntEntry)?.resume(index)
+            }
+        }
```
GregHib:
```suggestion
        continueDialogue("dialogue_pick_a_puppy:*") {
             (suspension as? Suspension.IntEntry)?.resume(index)
        }
        interfaceOption(id = "dialogue_pick_a_puppy:*") {
            (suspension as? Suspension.IntEntry)?.resume(index)
        }
```

game/src/main/kotlin/content/skill/summoning/pet/PetShopOwner.kt
```diff
@@ -0,0 +1,214 @@
...
+private val BREEDS = listOf(
+    Breed("Bulldog", "bulldog", "bulldog_puppy_white"),
+    Breed("Dalmatian", "dalmatian", "dalmatian_puppy_black"),
+    Breed("Greyhound", "greyhound", "greyhound_puppy_brown"),
+    Breed("Terrier", "terrier", "terrier_puppy_2"),
+    Breed("Sheepdog", "sheepdog", "sheepdog_puppy_black"),
+    Breed("Labrador", "labrador", "labrador_puppy_yellow"),
+)
+
+private val BREED_INDEX = BREEDS.withIndex().associate { (i, breed) -> breed.petId to i }
```
GregHib: See below
```suggestion
```

game/src/main/kotlin/content/skill/summoning/Summoning.kt
```diff
@@ -212,37 +224,62 @@ class Summoning : Script {
...
+                    else -> choice("Are you sure you want to dismiss your familiar?") {
                         option("Yes.") {
                             dismissFamiliar()
                         }
                         option("No.")
                     }
                 }
-                "Dismiss Now" -> dismissFamiliar()
             }
         }
```
GregHib, line 283: Might as well keep these separate

game/src/main/kotlin/content/skill/summoning/Summoning.kt
```diff
@@ -212,37 +224,62 @@ class Summoning : Script {
...
             renewFamiliar()
         }
 
-        interfaceOption("Call Follower", "*_details:call") {
-            callFollower()
+        interfaceOption("*", "*_details:call") {
+            when {
+                follower != null -> callFollower()
+                pet != null -> callPet()
+            }
```
GregHib: Keep these separate too, more readable with the correct options and ids rather than "*"

GregHib (changes requested):

data/entity/npc/monster/draconic/dragon.drops.toml
HarleyGilpin: Done.

engine/src/main/kotlin/world/gregs/voidps/engine/data/definition/ColumnReader.kt
HarleyGilpin: Done. Also removed the registry mapping, `ColumnDouble` in `ColumnType.kt`, and the `double` / `doubleOrNull` helpers in `Tables.kt` and `RowDefinition.kt` so the build compiles.

engine/src/main/kotlin/world/gregs/voidps/engine/data/PlayerSave.kt
HarleyGilpin: Reverted via `git checkout main --`.

game/src/main/kotlin/content/skill/summoning/pet/Incubator.kt
HarleyGilpin: Switched the end clock to `epochSeconds()` base. The `playerSpawn` block no longer resets the clock, it just restarts the poll timer when an incubation is in flight.

game/src/main/kotlin/content/skill/summoning/pet/PetShopOwner.kt
HarleyGilpin: Done. `continueDialogue` and `interfaceOption` now share a single `dialogue_pick_a_puppy:*` pair with the breed resolved from the matched id.

game/src/main/kotlin/content/skill/summoning/pet/PetShopOwner.kt
HarleyGilpin: Removed. `BREEDS.indexOfFirst` is used inline in the wildcard handler.

game/src/main/kotlin/content/skill/summoning/Summoning.kt
HarleyGilpin: Split into separate `familiar_details:dismiss` and `pet_details:dismiss` handlers.

game/src/main/kotlin/content/skill/summoning/Summoning.kt
HarleyGilpin: Split into separate `familiar_details:call` and `pet_details:call` handlers.

data/area/kandarin/sinclair_mansion/sinclair_mansion.areas.toml
```diff
@@ -1,3 +1,10 @@
 [sinclair_mansion_fishing_area]
 x = [2714, 2728]
 y = [3524, 3533]
+
+[sinclair_mansion_woods]
+x = [2680, 2740]
+y = [3470, 3540]
+level = 0
+hint = "in the woods near Sinclair Mansion."
```
GregHib:
```suggestion
```

Don't want to make it a penguin area

data/area/misthalin/varrock/grand_exchange/grand_exchange.areas.toml
```diff
@@ -0,0 +1,6 @@
+[grand_exchange]
+x = [3145, 3189]
+y = [3469, 3513]
+level = 0
+hint = "in the Grand Exchange."
```
GregHib:
```suggestion
```

engine/src/main/kotlin/world/gregs/voidps/engine/queue/ActionQueue.kt
```diff
@@ -109,21 +110,40 @@ class ActionQueue<C : Character>(
...
+                    true
+                }
+                is Suspension.Delay -> {
+                    suspension.resume()
+                    true
+                }
+                // Force-resuming a Custom whose predicate is still false makes
+                // the awaiting code believe its condition was met (e.g. that a
+                // dialogue closed when it did not). Skip when not ready and
+                // let the cleared queue drop the action.
```
GregHib: That's kind of the purpose of a long queue, on logout it executes all code even if it's not ready

game/src/main/kotlin/content/entity/player/dialogue/type/NPCDialogue.kt
```diff
@@ -46,6 +46,33 @@ suspend fun Player.npc(npcId: String, expression: String, text: String, largeHea
...
+ */
+private fun Player.splitDialogueLines(text: String): List<String> {
+    val font = get<FontDefinitions>().get("q8_full")
+    return if (text.contains("\n")) {
+        text.trimIndent().lines().flatMap { chunk -> font.splitLines(chunk, 400) }
+    } else {
+        font.splitLines(text, 400)
+    }
+}
+
```
GregHib: This can be reverted, `<br>` is fine, \n isn't necessary

game/src/main/kotlin/content/entity/player/dialogue/type/PlayerDialogue.kt
```diff
@@ -20,7 +20,12 @@ suspend inline fun <reified E : Expression> Player.player(text: String, largeHea
 
 @JvmName("playerExpression")
 suspend fun Player.player(expression: String, text: String, largeHead: Boolean = false, clickToContinue: Boolean = true, title: String? = null) {
-    val lines = if (text.contains("\n")) text.trimIndent().lines() else get<FontDefinitions>().get("q8_full").splitLines(text, 380)
+    val font = get<FontDefinitions>().get("q8_full")
+    val lines = if (text.contains("\n")) {
+        text.trimIndent().lines().flatMap { chunk -> font.splitLines(chunk, 400) }
+    } else {
+        font.splitLines(text, 400)
+    }
```
GregHib: Can also be reverted

game/src/main/kotlin/content/skill/summoning/pet/DogTalk.kt
```diff
@@ -0,0 +1,93 @@
...
+ * Inventory-conditional rows (Dalmatian + logs, Bulldog + cup_of_tea, Sheepdog + wool)
+ * win over the random pool when the relevant item is on the player.
+ */
+suspend fun Player.talkToDog(row: RowDefinition, dog: NPC) {
+    val breed = row.dogBreed() ?: return
+    val stageKey = if (row.stageForNpc(dog.id) == PetStage.Baby) "baby" else "grown"
+
+    val rows = Tables.get("pet_talks").rows().filter {
+        val stages = it.string("stage")
+        it.string("pet") == breed && (stages.isEmpty() || stages.split(',').any { s -> s.trim() == stageKey })
```
GregHib: Stages could be converted to "list<string>" and accessed with stringList rather than splitting by ',' at runtime

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,326 @@
...
+
+class KittenInteract : Script {
+
+    init {
+        val registered = mutableSetOf<String>()
+        for (row in allPetRows()) {
+            if (!row.isCatLike()) continue
+            for (npcId in listOfNotNull(row.npcOrNull("baby_npc"), row.npcOrNull("grown_npc"), row.npcOrNull("overgrown_npc"))) {
+                if (!registered.add(npcId)) continue
+                npcOperate("Interact-with", npcId) { interact ->
```
GregHib: As before, let's combine the npcIds into a list and have one npcOperate rather than calling it a whole bunch

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,326 @@
...
+            if (questCompleted("icthlarins_little_helper")) {
+                option("Did you understand what went on in that quest with the devourer and Icthlarin?") {
+                    icthlarinRecap(cat)
+                }
+            }
+            option("That's enough talking for now.") {
+                done = true
+            }
+        }
+    }
```
GregHib:
```suggestion
    fun menu() {
        choice("What would you like to ask?") {
            option("How are you doing?") {
                player<Quiz>("How are you doing?")
                npc<Happy>(cat.id, "I'm good. But could we go adventuring soon? I'm tired of talkin, meeoow.")
                menu()
            }
            option("How old are you now?") {
                player<Quiz>("How old are you now?")
                npc<Happy>(cat.id, "I'm not too old, and not too young. In fact I think I'm just right.")
                menu()
            }
            option("Where do you want to go?") {
                player<Quiz>("Where do you want to go?")
                npc<Happy>(cat.id, "Can we go to Varrock Sewer and chase some rats?")
                menu()
            }
            option("What do you want to do now?") {
                player<Quiz>("What do you want to do now?")
                npc<Happy>(cat.id, "I want to go chase things, kill them and then eat them. Purrr.")
                 menu()
            }
            if (questCompleted("icthlarins_little_helper")) {
                option("Did you understand what went on in that quest with the devourer and Icthlarin?") {
                    icthlarinRecap(cat)
                    menu()
                }
            }
            option("That's enough talking for now.")
        }
    }
```

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,326 @@
...
+        player<Quiz>("Aw come on. Just tell me one more time.")
+        npc<Happy>(cat.id, "Look, can we just leave it. Ask one of your adventuring friends.")
+        player<Happy>("Alright puss.")
+        return
+    }
+    player<Quiz>("Did you understand what went on in that quest with the devourer and Icthlarin?")
+    npc<Happy>(cat.id, "It wasn't all that difficult to understand.")
+    player<Quiz>("Really? I was, no actually am, still confused.")
+    npc<Happy>(cat.id, "Ok, I'll tell you what I think happened.")
+    player<Happy>("Great! Story time.")
```
GregHib, line 237: All `(cat.id,` can be removed if talkWith(npc) is called
```suggestion
private suspend fun Player.icthlarinRecap(cat: NPC) {
    talkWith(cat)
    val asked = inc("cat_icthlarin_asks", 1)
    if (asked >= 4) {
        player<Quiz>("Can you explain to me about Icthlarin and the devourer again?")
        npc<Happy>("I'm getting a little bit sick of talking about Icthlarin and all that stuff. If you didn't understand it the first time around, nor when I filled you in on a few details and then when I spelt it out to you what happened I don't think you'll ever get it.")
        player<Quiz>("Aw come on. Just tell me one more time.")
        npc<Happy>("Look, can we just leave it. Ask one of your adventuring friends.")
        player<Happy>("Alright puss.")
        return
    }
    player<Quiz>("Did you understand what went on in that quest with the devourer and Icthlarin?")
    npc<Happy>(cat.id, "It wasn't all that difficult to understand.")
    player<Quiz>("Really? I was, no actually am, still confused.")
    npc<Happy>(cat.id, "Ok, I'll tell you what I think happened.")
    player<Happy>("Great! Story time.")
```

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
```diff
@@ -0,0 +1,326 @@
...
+            }
+            option("What do you want to do now?") {
+                player<Quiz>("What do you want to do now?")
+                npc<Happy>(cat.id, "Curling up on a nice rug in front of a fire...")
+            }
+            option("That's enough talking for now.") {
+                done = true
+            }
+        }
+    }
```
GregHib: No loops!
```suggestion
        choice("What would you like to ask?") {
            option("How are you doing?") {
                player<Quiz>("How are you doing?")
                npc<Happy>(cat.id, "I'm as happy as a demon in a lava pit.")
                talkToHellcatWithAmulet(cat)
            }
            option("How old are you now?") {
                player<Quiz>("How old are you now?")
                npc<Happy>(cat.id, "I'm feeling a bit like a ghost in a cake shop.")
                player<Quiz>("A ghost in a cake shop? What do you mean?")
                npc<Happy>(cat.id, "You know, in need of exorcise.")
                talkToHellcatWithAmulet(cat)
            }
            option("Where do you want to go?") {
                player<Quiz>("Where do you want to go?")
                npc<Happy>(cat.id, "Let's go steal some things from stalls...")
                talkToHellcatWithAmulet(cat)
            }
            option("What do you want to do now?") {
                player<Quiz>("What do you want to do now?")
                npc<Happy>(cat.id, "Curling up on a nice rug in front of a fire...")
                talkToHellcatWithAmulet(cat)
            }
            option("That's enough talking for now.")
        }
```

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
```diff
@@ -0,0 +1,398 @@
...
+        return false
+    }
+    val spawned = try {
+        NPCs.add(npcStringId, tile)
+    } catch (t: Throwable) {
+        // Roll back the consumed item so a spawn failure (unknown id, NPC
+        // cap, world full, ...) doesn't silently destroy the player's pet.
+        if (!restart) inventory.add(itemId)
+        throw t
+    }
```
GregHib:
```suggestion
    val spawned = NPCs.add(npcStringId, tile)
```

NPCs.add can't throw anything

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
```diff
@@ -0,0 +1,398 @@
...
+ * Splits `Bark! (Translation!)` into `Bark!\n(Translation!)` so the chathead
+ * renders the bark on one line and the bracketed translation on the next.
+ */
+private fun breakParenTranslation(line: String): String {
+    val parenStart = line.indexOf('(')
+    if (parenStart <= 0) return line
+    val bark = line.substring(0, parenStart).trimEnd()
+    if (bark.isEmpty()) return line
+    val translation = line.substring(parenStart)
+    return "$bark\n$translation"
```
GregHib:
```suggestion
    return "$bark<br>$translation"
```

game/src/main/kotlin/content/skill/summoning/pet/Pets.kt
```diff
@@ -0,0 +1,167 @@
...
+
+private val npcIndex: Map<String, RowDefinition> by lazy {
+    buildMap {
+        for (row in Tables.get("pets").rows()) {
+            row.npcOrNull("baby_npc")?.let { put(it, row) }
+            row.npcOrNull("grown_npc")?.let { put(it, row) }
+            row.npcOrNull("overgrown_npc")?.let { put(it, row) }
+        }
+    }
+}
```
GregHib: 😕 

Just use Tables.itemOrNull("pets.${id}.baby_item") etc...
These maps can be removed entirely

game/src/main/kotlin/content/skill/summoning/pet/PetShopOwner.kt
```diff
@@ -0,0 +1,227 @@
...
+ *   .labrador (id 8)        -> index 5
+ */
+private val BREEDS = listOf(
+    Breed("Bulldog", "bulldog", "bulldog_puppy_white"),
+    Breed("Dalmatian", "dalmatian", "dalmatian_puppy_black"),
+    Breed("Greyhound", "greyhound", "greyhound_puppy_brown"),
+    Breed("Terrier", "terrier", "terrier_puppy_2"),
+    Breed("Sheepdog", "sheepdog", "sheepdog_puppy_black"),
+    Breed("Labrador", "labrador", "labrador_puppy_yellow"),
+)
```
GregHib: These could be put in a [dog_breeds] table and also removed along with `const val DOG_BREEDS: Set<String>`

GregHib (changes requested):

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
HarleyGilpin: Done — `breakParenTranslation` (and the dog-bark renderer) now return `"$bark<br>$translation"`, relying on the existing `<br>` chathead break.

Fixed in https://github.com/HarleyGilpin/void/commit/3a3daf97fa5b8f4771e91953677fff5751251960

game/src/main/kotlin/content/entity/player/dialogue/type/NPCDialogue.kt
HarleyGilpin: Reverted — NPC dialogue restored to main's line handling. The new `npc(npcId, animationId, text)` chathead-animation overload (needed by the pet talkers) is kept, and the wrap-width test expectations were restored too.

Fixed in https://github.com/HarleyGilpin/void/commit/a9ce0348899ff322eb2c51e9830cc7dc0a39e946

game/src/main/kotlin/content/entity/player/dialogue/type/PlayerDialogue.kt
HarleyGilpin: Reverted — player dialogue restored to main's line handling.

Fixed in https://github.com/HarleyGilpin/void/commit/9e1538f8eec90587b09beedeea17b24241c27028

engine/src/main/kotlin/world/gregs/voidps/engine/queue/ActionQueue.kt
HarleyGilpin: Reverted — the long queue now force-resumes Custom suspensions on logout regardless of readiness, like Continue/Delay, matching main.

Fixed in https://github.com/HarleyGilpin/void/commit/a0f58ecbeade71273528d427e6c5f8bbf470e1ed

data/area/kandarin/sinclair_mansion/sinclair_mansion.areas.toml
HarleyGilpin: Removed the `hint` line so the woods area no longer registers as a Penguin Hide-and-Seek location.

Fixed in https://github.com/HarleyGilpin/void/commit/ce1a9f6bbaa457e8832f5d3e97e1162b27ebd009

data/area/misthalin/varrock/grand_exchange/grand_exchange.areas.toml
HarleyGilpin: Removed the `hint` line so the Grand Exchange area no longer registers as a Penguin Hide-and-Seek location.

Fixed in https://github.com/HarleyGilpin/void/commit/a802d38858db99383ada56490e74b0c6718c7b18

game/src/main/kotlin/content/skill/summoning/pet/DogTalk.kt
HarleyGilpin: `stage` is now declared `list<string>` and read with `stringList`; the runtime `split(',')` is gone. Both readers (DogTalk and talkToPet) and every data row were updated.

Fixed in https://github.com/HarleyGilpin/void/commit/91ee87d407558834748c59f45f2ae42c3fb97752

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: Collapsed to a single `npcOperate("Interact-with", <csv of cat npc ids>)`, resolving the target inside the handler instead of registering one per id.

Fixed in https://github.com/HarleyGilpin/void/commit/b92bd649e50154f20da7532a3ea5e8a49b0955dd

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: Applied — `talkToCatWithAmulet` now uses a recursive local `menu()` and the quit option falls through; the while-loop is gone.

Fixed in https://github.com/HarleyGilpin/void/commit/4a36f14b5dda1d73023053e39db37e601b0cb108

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: Applied — `talkWith(cat)` is called once up front and every line uses the id-less `npc()` overload, so all the `cat.id` args are gone.

Fixed in https://github.com/HarleyGilpin/void/commit/451d8eff69d36c0a81c3efa206e3d35fa48d5875

game/src/main/kotlin/content/skill/summoning/pet/KittenInteract.kt
HarleyGilpin: Applied — the hellcat menu now re-opens via recursive `talkToHellcatWithAmulet(cat)` calls; no loop.

Fixed in https://github.com/HarleyGilpin/void/commit/807cec6e35d217c1f6c9aaac69401d34a26f62af

game/src/main/kotlin/content/skill/summoning/pet/Pet.kt
HarleyGilpin: Done — `val spawned = NPCs.add(npcStringId, tile)`; the try/catch rollback is removed.

Fixed in https://github.com/HarleyGilpin/void/commit/4e49aaf0c4aaa0f6a6cfc541ace73a3f6c8d0072

game/src/main/kotlin/content/skill/summoning/pet/Pets.kt
HarleyGilpin: Dropped the cached `itemIndex`/`npcIndex` maps — `petRowForItem`/`petRowForNpc` now scan `Tables.get("pets").rows()` directly. Callers only ever have the item/npc id, so it's a reverse scan over the table rather than a forward `Tables.itemOrNull` lookup.

Fixed in https://github.com/HarleyGilpin/void/commit/2ed132c8c1b1a97cf235803bc5aea537c584cc3c

game/src/main/kotlin/content/skill/summoning/pet/PetShopOwner.kt
HarleyGilpin: Moved the breeds into a `[dog_breeds]` data table (option / pet_id / puppy_item); removed the in-code `BREEDS` list and the `DOG_BREEDS` set, with `dogBreed()` and the shop reading the table.

Fixed in https://github.com/HarleyGilpin/void/commit/44c316a906fced3cb26c1306c6c1fec0f73dc21b

PR #987, opened 2026-05-18
Feat: Priest in Peril & Natures Spirit Quests

Ebp90:
The Priest in Peril quest.


Ebp90:
Issues:

Ghast's don't function properly.

1. They should put you in combat when they attempt rot your foot
2. They should be able to transform if you have a Druid Pouch with items in it
3. They don't fight back after transformed
4. If you use a pouch on them, after they transform you get the message "Nothing interesting happens"

The grotto bridge jump exact move looks off based on the data I have

An NPC despawn timer needs to be added for Filliman

game/src/main/kotlin/content/area/misthalin/paterdomus/TempleDoor.kt
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+    }
+
+    // ===== Per-stage dialogues (the two voices: Drezel = red, prompter = blue) =====
+
+    private suspend fun Player.droneVoicesProgress3() {
+        statement(" You knock at the door...<br>" +
+                "You hear a voice from inside.<br>" +
+                "<col=000080>You again? What do you want now?")
```
GregHib:
```suggestion
                "<navy>You again? What do you want now?")
```
If you wanted to

PR #996, opened 2026-05-28
fix: max roll for general_seeds table

advnau:
The current code was checking the combat level of the killer when passing a maximum roll to the loot method. This should be based off the npcs combat level instead.

Source: https://oldschool.runescape.wiki/w/Drop_table#General%20seed%20drop%20table


GregHib (approved):
Thanks for the fix!

PR #998, opened 2026-05-28
Test Gunnar's Ground happy path and missing req path + cutscene fix

Tect0r:
Fix Gunnar's Ground final cutscene progression

This fixes an issue where the final cutscene in Gunnar's Ground did not load, which prevented the player from progressing the quest.

I am new to this project, so feedback on whether this is the preferred approach would be appreciated.

Also adds tests for:

* the happy path
* the path where the player is missing the level 5 Crafting requirement

Resolves one task from #746.

Uses dragbone's helper functions for skipping dialogue and selecting dialogue options.


game/src/main/kotlin/content/area/misthalin/barbarian_village/Gudrun.kt
```diff
@@ -228,6 +221,13 @@ class Gudrun : Script {
         delay(2)
         gudrunHugging.anim("gudrun_hugging")
         delay(4)
+        cutscene.onEnd {
+            open("fade_out")
+            delay(3)
+            tele(3081, 3416)
+            clearCamera()
+            clearAnim()
+        }
```
GregHib: cutscene.onEnd triggers not only at the end but if the player exits another way (logout/x-log/walk) so it does still need to be set at the start of the cutscene. What was the bug you were seeing, it exiting too early?

Tect0r: hm than thats not a fix. 

When the cutscene starts it does the first fade out, then fades in, then ends the cutscene and does everything its told in the onEnd(). So fades out, delay, teleport. you then stand infront of Gudrun and cant progress the quest, since it never really starts the cutscene and never comes to the point where the next queststep is set.
it feels like the onEnd triggers the end of the cutscene aswell, even tho thats most likely not it.

Tect0r:
Fixed the cutscene ending abruptly by setting the `initialDelay` of the long queue in `onEnd` to `MAX_VALUE`.

Added a `walkTrigger` in `onEnd` so the player can cancel the cutscene by clicking a walkable tile.

GregHib (approved):

PR #1014, opened 2026-06-06
Fix summoning obelisk Renew-points interaction

maatheusgois-dd:
## Summary

- Add missing `Renew-points` object handler for summoning obelisks
- Restore drained summoning points to the player's max level, matching prayer altar behaviour
- Use a wildcard object match (same as `Infuse-pouch`) so small obelisk variants work without per-id registration
- Add `SummoningObeliskTest` coverage for restore and already-full cases

## Test plan

- [x] `./gradlew :game:test --tests content.skill.summoning.SummoningObeliskTest`
- [x] In-game: right-click obelisk → Renew-points restores summoning points after draining them
- [x] In-game: full summoning points shows "You already have full summoning points."

Made with [Cursor](https://cursor.com)

### DIFF
--- a/game/src/main/kotlin/content/skill/summoning/Summoning.kt
+++ b/game/src/main/kotlin/content/skill/summoning/Summoning.kt
@@ -145,6 +145,19 @@ fun Player.callFollower() {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


GregHib (approved):
Thanks for the contribution!

PR #1015, opened 2026-06-06
Implement beast of burden storage for summoning familiars

maatheusgois-dd:
## Summary

- Implements **Store** / **Take BoB** for beast-of-burden familiars (e.g. pack yak): withdraw/store items via `beast_of_burden` (671) and `summoning_side` (722) interfaces.
- Registers the `beast_of_burden` inventory (id 530, 6×5) and wires examine handlers.
- On dismiss, drops stored BoB items to the ground and closes the BoB interface.
- Adds `familiar_leash` timer and improves `Follow` teleport so familiars auto-call when they drift >15 tiles (skipped while in combat).
- Removes the "Not currently implemented" Take BoB stub.

Screenshot 2026-06-06 at 17 52 07

## Test plan

- [x] `BeastOfBurdenStoreTest` — store/withdraw round-trip
- [x] `BeastOfBurdenZeroSizeTest` — empty undersized inv recreated correctly
- [x] `BeastOfBurdenDisplayTest` — interface sync after store
- [x] `PackYakDefinitionTest` — pack yak has BoB capacity from NPC def
- [ ] Manual: summon pack yak → Interact → Store items → withdraw → Take BoB → dismiss (items drop)
- [ ] Manual: walk familiar >15 tiles away → familiar teleports back

Made with [Cursor](https://cursor.com)

### DIFF
--- a/game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
+++ b/game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
@@ -1,0 +1,221 @@
+package content.skill.summoning
+
+import content.entity.player.dialogue.type.intEntry
+import content.entity.player.modal.Tab
+import content.entity.player.modal.tab
+import world.gregs.voidps.engine.Script
+import world.gregs.voidps.engine.client.message
+import world.gregs.voidps.engine.client.sendScript
+import world.gregs.voidps.engine.client.ui.close
+import world.gregs.voidps.engine.client.ui.open
+import world.gregs.voidps.engine.data.definition.NPCDefinitions
+import world.gregs.voidps.engine.entity.character.player.Player
+import world.gregs.voidps.engine.entity.character.player.chat.inventoryFull
+import world.gregs.voidps.engine.entity.item.floor.FloorItems
+import world.gregs.voidps.engine.inv.beastOfBurden
+import world.gregs.voidps.engine.inv.clear
+import world.gregs.voidps.engine.inv.inventory
+import world.gregs.voidps.engine.inv.moveAll
+import world.gregs.voidps.engine.inv.sendInventory
+import world.gregs.voidps.engine.inv.transact.TransactionError
+import world.gregs.voidps.engine.inv.transact.operation.MoveItemLimit.moveToLimit
+
+private fun Player.familiarDef() = follower?.let { NPCDefinitions.get(it.id) }
+
+val Player.beastOfBurdenCapacity: Int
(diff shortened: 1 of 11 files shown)


game/src/main/kotlin/content/skill/summoning/SummoningTimers.kt
```diff
@@ -38,6 +38,18 @@ class SummoningTimers : Script {
...
+
+        timerTick("familiar_leash") {
+            ensureFollowerNearby()
+            if (follower == null) {
+                return@timerTick Timer.CANCEL
+            }
+            return@timerTick Timer.CONTINUE
+        }
+
+        timerStop("familiar_leash") {}
```
GregHib, line 51: This seems redundant if a familiar not in combat always has `Follow` mode which will keep them within range?

game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
```diff
@@ -0,0 +1,221 @@
...
+    val capacity = beastOfBurdenCapacity
+    if (capacity <= 0) {
+        return
+    }
+    if (beastOfBurden.size >= capacity) {
+        return
+    }
+    if (beastOfBurden.isEmpty()) {
+        inventories.clear("beast_of_burden")
+    }
```
GregHib, line 40: Is clearing the inventory necessary if it's already empty?

game/src/main/kotlin/content/skill/summoning/SummoningTimers.kt
HarleyGilpin: Addressed in #1042 — removed the `familiar_leash` timer entirely (and the `ensureFollowerNearby` helper it drove). You're right: `Follow.tick()` runs every tick and already teleports the familiar back when it drifts >15 tiles, and a familiar stays in `Follow` while not in combat, so the timer was redundant. Kept the `Follow` teleport improvement and the `callFollower` mode-reset.

game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
HarleyGilpin: Addressed in #1042. Clarified the intent: `inventories.clear("beast_of_burden")` removes the instance so the engine recreates it at the definition size on next access — it's a resize for a stale/undersized restored instance, not an item-wipe. The empty check is a safety guard so stored items are never discarded. Added a comment and tightened the condition.

PR #1021, opened 2026-06-10
feat: add bank pin management system

HarleyGilpin:
## Description

Implements the bank PIN system: a 4-digit PIN that protects bank and collection box access, managed through the  Bank PIN Settings interface (14) and PIN keypad interface (13).

## Manual Test Walkthrough
[Screencast_20260609_162129.webm](https://github.com/user-attachments/assets/6e0e38ae-c12e-428b-9583-0a587dfcb15a)

### PIN entry keypad (interface 13)
- Server-side digit shuffle packed into varps 562/563 (digits 4 bits each, entry stage in bits 26+), with the ten digit cells randomly repositioned after every press
- Buttons map back to digits through the per-player shuffled layout, so clicks can't be replayed positionally
- "I don't know it" (component 29) is labelled and unlocked at runtime — clicking it schedules the PIN for deletion after the recovery delay so a forgotten PIN never permanently locks an account
- Deletion banner (component 31) shows "YOUR PIN WILL BE DELETED IN X DAYS" while a deletion is pending

### PIN settings (interface 14)
- Set / Change / Delete PIN with confirmation overlay and authentic messages panel text
- Status display and per-state button visibility (no PIN / PIN set / change or deletion pending)
- Recovery delay toggle (3 ⇄ 7 days)
- "Cancel the PIN that's pending" aborts a pending change or deletion


game/src/main/kotlin/content/entity/player/bank/pin/BankPinSettings.kt
```diff
@@ -0,0 +1,279 @@
...
+                this,
+                "set",
+                "Do you really wish to set a PIN on your bank account?",
+                "Yes, I really want a Bank PIN. I will never forget it!",
+                "No, I might forget it!",
+            )
+        }
+
+        interfaceOption("Change your recovery delay", "bank_pin_settings:recovery*") {
+            set("bank_pin_long_recovery", !get("bank_pin_long_recovery", false))
```
GregHib:
```suggestion
            toggle("bank_pin_long_recovery")
```

GregHib (approved):

PR #1024, opened 2026-06-10
feat: scheduled reload of account definitions from storage

HarleyGilpin:
## Summary 
Website password resets and imported accounts previously required a server restart because AccountDefinitions caches all account metadata (including password hashes) at startup and never refreshes.

- Add AccountDefinitions.merge(): add/update only, mutates existing definitions and clans in place (preserves live Clan.members), skips online players and accounts with pending saves whose in-memory state may be newer than storage.
- Add AccountDefinitionsReloader: loads names/clans on Dispatchers.IO, applies merge on the game thread; SupervisorJob so a failed reload does not cancel future ones.
- Add AccountReload script: world timer driven by new setting storage.accounts.reloadMinutes (-1 default, disabled), reacts to ::reload settings.
- Add ::reload accounts admin command for on-demand refresh.
- FileStorage.names()/clans(): warn and skip corrupt save files instead of crashing, so a partially-written save cannot abort a reload.
- Swap AccountDefinitions maps to ConcurrentHashMap as login threads read while the game thread merges.

Known limitation: a password reset for a currently online player is overwritten at logout; it converges on the next reload after they log out.

## Tests successful
- engine/src/test/kotlin/world/gregs/voidps/engine/data/definition/AccountDefinitionsTest.kt
- engine/src/test/kotlin/world/gregs/voidps/engine/data/AccountDefinitionsReloaderTest.kt

## Manual test
- Successfully reloaded password for a test account via in-game 'reload accounts' command.

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/data/definition/AccountDefinitions.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/data/definition/AccountDefinitions.kt
@@ -1,6 +1,5 @@
 package world.gregs.voidps.engine.data.definition
 
-import it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap
 import world.gregs.voidps.engine.data.Storage
 import world.gregs.voidps.engine.data.config.AccountDefinition
 import world.gregs.voidps.engine.entity.character.player.Player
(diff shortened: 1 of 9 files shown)


engine/src/main/kotlin/world/gregs/voidps/engine/data/definition/AccountDefinitions.kt
```diff
@@ -70,4 +70,57 @@ class AccountDefinitions(
...
+                existing.name = clan.name
+                existing.friends = clan.friends
+                existing.ignores = clan.ignores
+                existing.joinRank = clan.joinRank
+                existing.talkRank = clan.talkRank
+                existing.kickRank = clan.kickRank
+                existing.lootRank = clan.lootRank
+                existing.coinShare = clan.coinShare
+            }
+        }
```
GregHib, line 123: Why not just `clans.putAll(clanUpdates)` and apply the filter before?

HarleyGilpin: `Clan` instances are live-referenced at runtime, so I avoided replacing them wholesale. Online members hold the instance directly (`player.clan` is set on join and only cleared on leave), and `Clan.members` is the live list of online members.

The `skip` filter only protects clans whose owner is online or mid-save, but a clan can have online members chatting while its owner is offline. For those, `putAll` would swap in a fresh instance with an empty `members` list, the current members would keep chatting through the orphaned old instance while new joiners get the new one, silently splitting the chat and breaking kicks/rank updates for the old group.

Mutating the fields in place keeps the instance identity stable so existing references stay valid, and the refreshed ranks/friends apply immediately to anyone already in the chat. Happy to restructure if you would still prefer a different shape here.

GregHib (approved):

PR #1040, opened 2026-06-21
Cleanup to Groats farm (12851)

M1krofin:
Added farmer combat def, anims, sounds and drop tables (f2p/p2p).
Added Haybale, sack, water barrel, crates functionality.
Added Seth Groats dialogue.
Fixed emptying bucket message to correct one.
Added ducks / drake combat defs, sounds, anims and droptable.
Added object examines.
Last and most important addition: Quack!


data/entity/npc/humanoid/farmer/farmer.drops.toml
```diff
@@ -0,0 +1,64 @@
...
+    { id = "coins", amount = 15, chance = 4 },
+    { id = "coins", amount = 25, chance = 1 },
+    { id = "nothing", amount = 0, chance = 8 },
+    { id = "fishing_bait", chance = 5 },
+    { id = "copper_ore", chance = 2 },
+    { id = "earth_talisman", chance = 2 },
+    { id = "cabbage", chance = 1  },
+    { id = "bronze_med_helm", chance = 2 },
+    { id = "iron_dagger" },
+]
```
GregHib: These can be combined and have `, members = false` or `, members = true` on the drops which are specific to f2p or members

M1krofin: Yep makes sense, now fixed. Thanks for the tip as always.

PR #1042, opened 2026-06-23
Implement beast of burden storage for summoning familiars

HarleyGilpin:
Re-lands #1015 (originally by @maatheusgois-dd), which was closed with two unaddressed review comments. Both are now resolved. I also added familiar storage compaction when player opens familiar storage, which cause familiar held items to move to the top of the familiar's inventory, occupying any empty slots.

## Summary

- Implements **Store** / **Take BoB** for beast-of-burden familiars (e.g. pack yak): withdraw/store via the `beast_of_burden` (671) and `summoning_side` (722) interfaces.
- Registers the `beast_of_burden` inventory (id 530, 6×5) and wires examine handlers.
- On dismiss, drops stored BoB items to the ground and closes the BoB interface.
- Improves `Follow` teleport so familiars auto-call when they drift >15 tiles (skipped while in combat).
- Removes the "Not currently implemented" Take BoB stub.
- Add BoB storage compaction (Consolidates items to the top of the familiar inventory) based on this source: https://www.youtube.com/watch?v=6--CWVKORGo

Comparison against the RuneScape wiki (https://runescape.wiki/w/Beasts_of_Burden?oldid=4736259) surfaced several behaviors that were missing or diverged:

- Only tradeable items can be stored. Reject untradeable/lent/dummy items
  in store(), mirroring the trade-screen restriction, with a message.
- The familiar inventory can't be accessed in combat. Gate
  openBeastOfBurden() on the under_attack clock.
- Dropped items are now owner-only and retrievable for five minutes before
  despawning (was: never despawn), and use the wiki's drop message.
- A familiar slain in combat now drops its stored items and is dismissed.
  Tag the familiar NPC with its owner on summon and handle its death via
  npcDeath, reusing dismissFamiliar (without re-removing the despawning NPC)
- A familiar can't carry an unstackable item worth more than 5m, nor a
stack whose total value exceeds 5m.
- Standard BoBs cannot carry rune essence/pure essence.
- Add essence only BoBs (Abyssal parasite/lurker/titan)

## Tests

`./gradlew :game:test --tests "content.skill.summoning.BeastOfBurden*" --tests "content.skill.summoning.PackYakDefinitionTest"` — all pass:

- `BeastOfBurdenStoreTest` — store/withdraw round-trip
- `BeastOfBurdenZeroSizeTest` — empty undersized inv recreated correctly
- `BeastOfBurdenDisplayTest` — interface sync after store
- `PackYakDefinitionTest` — pack yak BoB capacity from NPC def


game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
```diff
@@ -0,0 +1,217 @@
...
+        if (usedSlots >= capacity && bob.indexOf(item.id) == -1) {
+            player.message("Your familiar can't carry any more items.")
+            return
+        }
+        player.inventory.transaction {
+            val moved = moveToLimit(item.id, amount, player.beastOfBurden)
+            if (moved == 0) {
+                error = TransactionError.Full()
+            }
+        }
```
GregHib: Worth noting the original PR had an issue where depositing more items than in inv would somehow mean the next lot of deposited items skips some slots.

HarleyGilpin: Good catch — fixed in d2aa41f58. Root cause: when the requested amount exceeds what's actually held, `MoveItemLimit.moveToLimit` hits its undo path (`removed < added`), which removes the first `added` matches from the target (including items already stored) and re-adds only `removed`, scattering the survivors into high slots and leaving gaps, so the next deposit fills the gaps and looks like it skips slots.

Clamped the requested amount to the held count before `moveToLimit` in both store and withdraw, so `removed == added` and the undo path never runs. Added `BeastOfBurdenStoreLimitTest` covering both directions (it reproduces the gap on the old code: expected first-empty-slot 8, got 5).

game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
```diff
@@ -0,0 +1,304 @@
...
+import world.gregs.voidps.engine.inv.inventory
+import world.gregs.voidps.engine.inv.sendInventory
+import world.gregs.voidps.engine.inv.transact.TransactionError
+import world.gregs.voidps.engine.inv.transact.operation.MoveItemLimit.moveAllToLimit
+import world.gregs.voidps.engine.inv.transact.operation.MoveItemLimit.moveToLimit
+import world.gregs.voidps.engine.timer.toTicks
+import java.util.concurrent.TimeUnit
+
+/** Items (or stacks) worth more than this can't be carried by a familiar. */
+private const val MAX_BEAST_OF_BURDEN_VALUE = 5_000_000L
```
GregHib: Can we add this to game.properties?

game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
```diff
@@ -0,0 +1,304 @@
...
+
+/** The only items the abyssal essence familiars carry; every other familiar refuses them. */
+private val BEAST_OF_BURDEN_ESSENCE = setOf("rune_essence", "pure_essence")
+
+private fun Player.familiarDef() = follower?.let { NPCDefinitions.get(it.id) }
```
GregHib: Can be inlined to `follower?.def`

GregHib (changes requested):
I think you should be able to use items on BoB familiars to store them? 🤔

game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
HarleyGilpin: Moved to `game.properties` as `summoning.beastOfBurden.maxValue=5000000` (read via a new `Settings` `Long` accessor). 86a38ecde

game/src/main/kotlin/content/skill/summoning/BeastOfBurden.kt
HarleyGilpin: Inlined to `follower?.def` - dropped the `familiarDef()` helper. 86a38ecde

HarleyGilpin:
> I think you should be able to use items on BoB familiars to store them? 🤔

Added. Using an item on a familiar now stores it via `itemOnNPCOperate("*", "*_familiar")`, reusing `store()` so ownership/capacity/value/tradeable/essence rules and the in-combat gate all apply. Covered by new tests in `BeastOfBurdenWikiTest`. 86a38ecde

GregHib (approved):

PR #1052, opened 2026-06-29
Summoning: complete familiar abilities (specials, passives, dialogue)

HarleyGilpin:
### Summary
Fleshes out summoning familiars: talk dialogue, special moves for every familiar, foraging, skill boosts, passive abilities and overhead chat.

### Talk dialogue & chatheads
- Per-familiar Interact talk dialogue for ~60 familiars (random + context-aware lines).
- Familiars use a large, static correct-model chathead.
- Interact (and every other option) keeps working on the wilderness `<id>_combat` form - transformed npcs fall back to their base id's interaction handlers.

### Special moves
Every familiar's cast-button special is implemented - combat hits, stat drains, stuns/binds, AoEs, boosts, heals, teleports, foraging drops and banking. Highlights:
- **Cast surfaces**: plain click (instant, or auto-fired at the familiar's current combat target as in the live game), click-a-target (npc/player/scenery), Cast-on-inventory-item (Swallow Whole, Immense Heat, Ophidian Incubation, Winter Storage) and Cast-on-ground-item (Rise from the Ashes). Item-target specials print a usage hint on a bare click instead of doing nothing.
- **Projectiles** fly at the live game's distance-scaled speed (1 tick + 1/3 tick per tile).

### Passive & other abilities
- Hunting cat teleports (graahk/kyatt/larupia; lava titan's Lava Maze option now gated + animated), bunyip raw-fish -> water runes and 20 LP/15s passive heal, geyser titan bowl-boiling and glory recharging, giant ent essence -> earth/nature runes (stacking), pyrelord/forge regent log-burning beneath themselves (with the player-style step-west), spirit dagannoth Ferocious double-strike, ibis/granite lobster harpoon substitution and fishing-gated foraging with a cut of the xp, karamthulhu Drown / smoke devil Flames attack orders, cockatrice Drain recast, unicorn Cure (fixed - it checked a nonexistent variable).
- Style-gated defence passives: iron/steel titan owners are 10%/15% harder to hit with melee, wolpertinger owners 5% harder with magic.
- Data-driven `Player.familiarBoost(skill)` now supports level-scaled formulas (stranger plant Farming +1+4%, geyser Ranged +1+3%) alongside flat boosts; the graahk's missing Hunter +5 restored.
- Familiars follow the real PvP rules when sent at players (wilderness or PvP area + combat-level bracket via `Target.attackable`), not a Clan-Wars-only flag.

### Foraging
- Generic forager system: gathers from a forage_<id> drop table into the familiar inventory every ~30s, retrieved via Withdraw/Take (reuses the beast-of-burden UI). Covers 20 overworld foragers; the fishing foragers only produce while their owner fishes.
### Engine fixes surfaced by this work
- A second same-tick graphic no longer overwrites the first (owner sparkles now render alongside the scroll-cast flourish).
- Familiar AoEs find npcs by their bodies, not just their anchor tile - large monsters beside the familiar count.
- NPC followers face an adjacent idle owner instead of shuffling onto the follow tile, and step out to a free tile when stood on (as combat movement does).
- Projectile model orientation: cache graphic patching (`GraphicEncoder` + `FixGraphics` tool, wired into `CacheBuilder`) rotates the backwards-facing Toad Bark cannonball 180 degrees - the protocol itself has no projectile rotation.

### Overhead dialogue
- Familiars shout a random overhead line ~every 30s; 83 familiars' lines sourced from the RS Wiki, stored as a data-driven familiar_overhead table.

### Manual Testing
Tracked per-familiar in the QA sheet (special move, boosts, other abilities, in-game visuals):
https://docs.google.com/spreadsheets/d/14LjVE4VdZ8FEGrl42_MJjrY3iAfAV_6CJ1ZuV31r-eQ/edit?usp=sharing

In-game verification during development caught and fixed: silent cast buttons, backwards projectiles, wrong attack animations, PvP-area refusals, non-stacking runes, forager output, and follow/step-out movement quirks.

### Unit Tests
`game/src/test/kotlin/content/skill/summoning` - all passing, including:
- `FamiliarSpecialMoveTest` - the scroll/points cast gate (spend-on-success, refusals)
- `FamiliarSpecialEffectTest`, `FamiliarCombatSpecialEffectTest`, `FamiliarUtilitySpecialEffectTest` - every special's effect through the registered blocks and, where the surface matters, the real dispatchers (orb cast clicks, item-on-npc, cast-on-item, dialogue options)
- Beast-of-burden, forager, heal, and definition tests
- Full `:engine:test` and `:game:test` suites green

### References
Special Moves (Familiars lvl 1-40): https://www.youtube.com/watch?v=kduBCqBiLB0
Summoning Scrolls Knowledge Base Article: https://2011.rs/kb/summoning_scrolls
darkan (primary mechanics reference): local checkout, `Scroll.java` / `Pouch.java`
2009scape (animation/config reference): local checkout, `npc_configs.json`


data/skill/summoning/summoning.anims.toml
```diff
@@ -1,38 +1,16 @@
...
-id = 8411
-
-[expression_leech_sad]
-id = 8412
-
-[expression_leech_neutral]
-id = 8413
-
-[expression_leech_angry]
-id = 8414
```
GregHib, line 38: Were these all incorrect?

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/interact/ItemOnNPCInteract.kt
```diff
@@ -15,9 +15,13 @@ data class ItemOnNPCInteract(
     val id: String,
     val player: Player,
 ) : Interact(player, target) {
-    override fun hasOperate() = Operation.itemOnNpc.containsKey("${item.id}:*") || Operation.itemOnNpc.containsKey("${item.id}:${target.def(player).stringId}") || Operation.itemOnNpc.containsKey("*:${target.def(player).stringId}")
+    // A transformed npc (e.g. a familiar's wilderness "<id>_combat" form) falls back to its base
+    // id's handlers when the transform has none of its own for the item.
+    override fun hasOperate() = Operation.itemOnNpc.containsKey("${item.id}:*") || Operation.itemOnNpc.containsKey("${item.id}:${target.def(player).stringId}") || Operation.itemOnNpc.containsKey("*:${target.def(player).stringId}") ||
+        Operation.itemOnNpc.containsKey("${item.id}:${target.id}") || Operation.itemOnNpc.containsKey("*:${target.id}")
```
GregHib: Using transformed ids is intentional and don't really want to start adding combinations, should use prefixes instead

game/src/main/kotlin/content/entity/combat/Target.kt
```diff
@@ -50,12 +54,18 @@ object Target {
...
+            // The "Attack" option is a player-facing menu concern. A familiar's base (non-wilderness)
+            // form deliberately has no "Attack" option so its owner can't click it, but an NPC must
+            // still be able to retaliate against a familiar that attacks it, so skip the check there.
+            val retaliatingAgainstFamiliar = source is NPC && target["owner_index", -1] != -1
+            if (!retaliatingAgainstFamiliar) {
+                if (target.transform != "") {
+                    if (!NPCDefinitions.get(target.transform).options.contains("Attack")) {
+                        return false
+                    }
+                } else if (target.def.options[1] != "Attack") {
```
GregHib: Can probably be applied globally no need for familiar check for this?
```suggestion
if (source is Player && target.def(source).options.contains("Attack") {
} else if (target.def.options.contains("Attack") {
}
```

game/src/main/kotlin/content/entity/death/NPCDeath.kt
```diff
@@ -56,7 +57,10 @@ class NPCDeath(
             val npc = this
             val onDeath = Death.killed(npc)
             queue(name = "death", 1) {
-                val killer = killer
+                // Credit a familiar's kill to its owner so loot, slayer and the kill log go to the
+                // player rather than the familiar npc.
+                val killer = (killer as? NPC)?.get("owner_index", -1)?.takeIf { it != -1 }
+                    ?.let { Players.indexed(it) } ?: killer
```
GregHib: Cleaner to read and understand imo
```suggestion
                var killer = killer
                if (killer is NPC && killer.contains("owner_index")) {
                    killer = Players.indexed(killer.getValue("owner_index"))
                }
```

game/src/main/kotlin/content/entity/player/dialogue/DialogueCommon.kt
```diff
@@ -19,11 +22,43 @@ fun Interfaces.sendChat(
...
+    }
+    val values = (VariableDefinitions.get("follower_details_chathead_animation")?.values as? MapValues)?.values ?: return null
+    val value = values[npcId.removeSuffix("_familiar")] ?: return null
+    return if (value > 50) {
+        EnumDefinitions.int("pet_details_chathead_animations_sad", value - 50)
+    } else {
+        EnumDefinitions.int("pet_details_chathead_animations_normal", value)
+    }
+}
+
```
GregHib, line 62: I'm pretty sure from the last PR that these are pet only animations, doesn't include familiars

game/src/main/kotlin/content/skill/summoning/EnchantedHeadgear.kt
```diff
@@ -0,0 +1,221 @@
...
+    Headgear("farseer_helm", "farseer_helm_enchanted", "farseer_helm_charged", 30, 70),
+    Headgear("rune_full_helm", "rune_full_helm_enchanted", "rune_full_helm_charged", 30, 60),
+    Headgear("splitbark_helm", "splitbark_helm_enchanted", "splitbark_helm_charged", 30, 50),
+    Headgear("helm_of_neitiznot", "helm_of_neitiznot_enchanted", "helm_of_neitiznot_charged", 45, 90),
+    Headgear("dragon_med_helm", "dragon_helm_enchanted", "dragon_helm_charged", 50, 110),
+    Headgear("lunar_helm", "lunar_helm_enchanted", "lunar_helm_charged", 55, 110),
+    Headgear("armadyl_helmet", "armadyl_helmet_enchanted", "armadyl_helmet_charged", 60, 120),
+)
+
+private val BY_ID = HEADGEAR.flatMap { listOf(it.base to it, it.enchanted to it, it.charged to it) }.toMap()
```
GregHib: You know the drill, put it in tables

game/src/main/kotlin/content/skill/summoning/FamiliarBoosts.kt
```diff
@@ -0,0 +1,66 @@
...
+    "desert_wyrm_familiar" to mapOf(Skill.Mining to Boost(1)),
+    "void_ravager_familiar" to mapOf(Skill.Mining to Boost(1)),
+    "obsidian_golem_familiar" to mapOf(Skill.Mining to Boost(7)),
+    "lava_titan_familiar" to mapOf(Skill.Mining to Boost(10), Skill.Firemaking to Boost(10)),
+    "pyrelord_familiar" to mapOf(Skill.Firemaking to Boost(3)),
+    "forge_regent_familiar" to mapOf(Skill.Firemaking to Boost(4)),
+    "magpie_familiar" to mapOf(Skill.Thieving to Boost(3)),
+    "stranger_plant_familiar" to mapOf(Skill.Farming to Boost(1, 0.04)),
+    "geyser_titan_familiar" to mapOf(Skill.Ranged to Boost(1, 0.03)),
+)
```
GregHib: Can also go in a table

game/src/main/kotlin/content/skill/summoning/FamiliarSpecialMoves.kt
```diff
@@ -0,0 +1,179 @@
...
+ *  - [floorItemTarget] : specials where the player then clicks a ground item (phoenix's ashes)
+ *
+ * Item-target specials (a fish/egg/bar) can also be triggered by using the item on the familiar,
+ * wired separately with `itemOnNPCOperate` in their own scripts (operate, not approach, so their
+ * exact item:npc key beats the beast-of-burden store's wildcard). Both surfaces run the same block.
+ *
+ * Each block returns whether the move actually happened so [castFamiliarSpecial] only spends a
+ * scroll + points on a real cast.
+ */
+object FamiliarSpecialMoves {
```
GregHib: Will also need a clear() which is called by Script.clear

GregHib:
1. Transforming npcs: Werewolf, kalphiteQueen, Shades use `if (source is Player) { interactPlayer(source, "Attack")` to retaliate, if the attacker is your familiar they'll need to attack the owner instead (currently they'll get stuck in PauseMode and be indefinitely unattackable) 

2. In single combat areas you can't attack the npc your familiar is attacking and visa-versa not sure if this is correct or not. And you can also get your familiar to attack one npc where you attack another.

3. Opening and closing world map (and other full screen interfaces) breaks/removes the right click options on the summoning orb - The settings will need to be resent on refresh/open

4. Special moves have no range limit, e.g. with vampyre bat you can start in lumbridge, tele to goblins and use an attack scroll and it'll do damage even though it's > 30 tiles away. (same with beaver and trees)

5. No combat sounds when familiars are fighting with npcs

game/src/main/kotlin/content/entity/player/dialogue/DialogueCommon.kt
GregHib: Hmm seems to work fine

game/src/main/kotlin/content/entity/player/dialogue/type/NPCDialogue.kt
```diff
@@ -58,7 +59,14 @@ private suspend fun Player.npc(lines: List<String>, clickToContinue: Boolean, np
...
     sendNPCHead(this, id, head, npcDef.id)
-    interfaces.sendChat(id, head, if (npcDef.contains("dialogue")) "${npcDef["dialogue", ""]}_$expression" else expression, title ?: npcDef.name, lines)
+    val chathead = familiarChatheadAnimation(npcId)
+    when {
+        // Familiar chatheads play the same cache animation as the familiar details interface,
+        // regardless of expression - dialogue expressions only exist for humanoid heads.
+        chathead != null -> interfaces.sendChat(id, head, chathead, title ?: npcDef.name, lines)
+        npcDef.contains("dialogue") -> interfaces.sendChat(id, head, "${npcDef["dialogue", ""]}_$expression", title ?: npcDef.name, lines)
+        else -> interfaces.sendChat(id, head, expression, title ?: npcDef.name, lines)
+    }
```
GregHib: This is too low level of a change, use `npc(npcId: String, expression: String, text: String` if custom expression needs to be passed

tools/src/main/kotlin/world/gregs/voidps/tools/cache/FixGraphics.kt
```diff
@@ -0,0 +1,65 @@
...
+import world.gregs.voidps.buffer.write.ArrayWriter
+import world.gregs.voidps.cache.CacheDelegate
+import world.gregs.voidps.cache.Index
+import world.gregs.voidps.cache.definition.data.GraphicDefinition
+import world.gregs.voidps.cache.definition.decoder.GraphicDecoder
+import world.gregs.voidps.cache.definition.encoder.GraphicEncoder
+
+object FixGraphics {
+
+    /** The barker toad's Toad Bark cannonball, whose model faces backwards in this revision. */
```
GregHib: I don't think this is the correct cannonball anim. See: https://youtu.be/o0zwPoONP8I?&t=17

Maybe it's 1401 for shoot gfx and 2143 for projectile (same as hand cannon)?

game/src/main/kotlin/content/skill/summoning/familiar/RavenousLocust.kt
```diff
@@ -0,0 +1,46 @@
+package content.skill.summoning.familiar
+
+import content.entity.player.dialogue.Happy
+import content.entity.player.dialogue.Neutral
+import content.entity.player.dialogue.type.npc
+import content.entity.player.dialogue.type.player
+import world.gregs.voidps.engine.Script
+import world.gregs.voidps.type.random
+
+class RavenousLocust : Script {
```
GregHib, line 10: Using Famine on an npc target causes the player to attack

game/src/main/kotlin/content/skill/summoning/familiar/Pyrelord.kt
```diff
@@ -0,0 +1,169 @@
...
+        itemOnNPCOperate("gold_bar", "pyrelord_familiar") { (npc, item) ->
+            if (npc != follower) {
+                return@itemOnNPCOperate
+            }
+            castFamiliarSpecial { immenseHeat(item.id) }
+        }
+
+        // The pyrelord and forge regent act as portable fire sources - the demon breathes fire over
+        // the logs, lighting them beneath itself, granting the log's xp plus a bonus for its help.
+        for (familiar in listOf("pyrelord_familiar", "forge_regent_familiar")) {
```
GregHib, line 94: forge_regent needs different animations

game/src/main/kotlin/content/skill/summoning/familiar/SpiritDagannoth.kt
```diff
@@ -0,0 +1,65 @@
...
+class SpiritDagannoth : Script {
+    init {
+        // Ferocious - about a fifth of the dagannoth's melee attacks lash out a second time.
+        npcCombatAttack("spirit_dagannoth_familiar") { attack ->
+            if (attack.type != "melee" || this["ferocious", false] || random.nextInt(FEROCIOUS_CHANCE) != 0) {
+                return@npcCombatAttack
+            }
+            // The extra hit re-fires this handler - the flag keeps it from chaining.
+            this["ferocious"] = true
+            anim("ferocious")
```
GregHib, line 26: This special is meant to stun the enemy?

game/src/main/kotlin/content/skill/summoning/familiar/IronTitan.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import content.entity.player.dialogue.Happy
+import content.entity.player.dialogue.Neutral
+import content.entity.player.dialogue.Quiz
+import content.entity.player.dialogue.type.npc
+import content.entity.player.dialogue.type.player
+import world.gregs.voidps.engine.Script
+import world.gregs.voidps.type.random
+
+class IronTitan : Script {
```
GregHib, line 11: Iron/steel titans attack target doesn't clear after target death so if a familiar respawns you can call it and it'll re-engage the same target as before.

GregHib (changes requested):
A few comments and fixes but overall looking really strong

data/skill/summoning/summoning.anims.toml
HarleyGilpin: Chathead expression logic for pets and familiars was moved. The follower_details_chathead_animation varbit map gives the familiar or pet's headchat animation value.

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/interact/ItemOnNPCInteract.kt
HarleyGilpin: Reverted the engine change; item-on-familiar handlers now register with a `<id>*` prefix so the wilderness `_combat` form's keys are expanded at registration instead: https://github.com/GregHib/void/pull/1052/commits/aba7ae8e1c83fccb20f8db5df71e6532c43fa277

game/src/main/kotlin/content/entity/combat/Target.kt
HarleyGilpin: Applied in https://github.com/GregHib/void/pull/1052/commits/3f1851510748391e2f842de2882adec798a7e7d8 - player sources now use `target.def(source)` (which already resolves `transform_id`). One caveat: `combat()` re-validates `attackable()` every tick, and a familiar's base form deliberately has no Attack option, so a fully global check would make npcs disengage from familiars mid-retaliation - npc sources keep a minimal `owner_index` guard for that.

game/src/main/kotlin/content/skill/summoning/EnchantedHeadgear.kt
HarleyGilpin: Moved to the `enchanted_headgear` table in `enchanted_headgear.tables.toml`: https://github.com/GregHib/void/pull/1052/commits/8ca083c80776a68bf99e95cfdce8b02168d945e9

game/src/main/kotlin/content/skill/summoning/FamiliarBoosts.kt
HarleyGilpin: Moved to the `familiar_boosts` table in `familiar_boosts.tables.toml` (flat amount + percentage of base level): https://github.com/GregHib/void/pull/1052/commits/14efc4605e8014aba58ae6c4c00dd7c185b0fe80

game/src/main/kotlin/content/skill/summoning/FamiliarSpecialMoves.kt
HarleyGilpin: Added - `FamiliarSpecialMoves` now implements `AutoCloseable` and is registered alongside `PrayerApi`/`RandomEvents` in `ContentLoader`, so `Script.clear` empties all six registries: https://github.com/GregHib/void/pull/1052/commits/c168982b20f8965ab5823183d82181b7c807c91b

game/src/main/kotlin/content/entity/player/dialogue/type/NPCDialogue.kt
HarleyGilpin: Reverted the renderer change. The dialogue-target overload (which already resolves the transform id) now resolves `familiarChatheadAnimation` and passes it through `npc(npcId, expression, text)` as the expression, with `sendChat` falling back to raw animation names/ids for such custom expressions - the pet talk script does the same: https://github.com/GregHib/void/pull/1052/commits/9ab8bf0789323b6cc1f8b13eb220a1806a599323

tools/src/main/kotlin/world/gregs/voidps/tools/cache/FixGraphics.kt
HarleyGilpin: You're right - matched the video with 1401 for the shoot gfx and the hand cannon's 2143 for the projectile, and dropped the whole cache rotation patch (FixGraphics/GraphicEncoder): https://github.com/GregHib/void/pull/1052/commits/a4d6c9d87bca2d4af92c15d2c4d7c130b42a9b57

game/src/main/kotlin/content/skill/summoning/familiar/Pyrelord.kt
HarleyGilpin: Gave it its own breath animation (7871, the same one its Inferno special uses; the pyrelord keeps 8085): https://github.com/GregHib/void/pull/1052/commits/96a5a9b3805380ed512699979fb88907d1f94feb

game/src/main/kotlin/content/skill/summoning/familiar/RavenousLocust.kt
HarleyGilpin: I couldn't reproduce this server-side - added a regression test sending the real `InteractInterfaceNPC` instruction for `familiar_details:cast_famine` at an npc, and the locust engages while the player never enters combat: https://github.com/GregHib/void/pull/1052/commits/0359919da1ce3fc9483807e19defc31343a6b3f0. Could you share the exact steps (details-tab cast vs orb cast, and whether you were already in combat)? I suspect the client routes the click differently than our component mapping expects (a few cast component ids are still unconfirmed), making the npc click fall through to a plain Attack.

game/src/main/kotlin/content/skill/summoning/familiar/SpiritDagannoth.kt
HarleyGilpin: This block is the Ferocious passive rather than the special - per the wiki the spirit dagannoth has "a chance to attack twice in quick succession, called Ferocious, similar to the honey badger's ability" (https://runescape.wiki/w/Spirit_dagannoth). The Spike Shot special does stun, implemented in FamiliarCombatSpecials (`spike_shot` + `stun(hit, SPIKE_SHOT_STUN_TICKS)`). No change made - happy to adjust if you'd rather it read differently.

game/src/main/kotlin/content/skill/summoning/familiar/IronTitan.kt
HarleyGilpin: Root cause: combat only emits `combatStop` for fights whose first swing landed (`started`), so a titan engaged mid-retaliation whose target died pre-swing kept the stale target and the called special would re-engage it after respawn. Death now clears every attacker's matching target, with a red/green regression test: https://github.com/GregHib/void/pull/1052/commits/bbe99806bb448b163cd1902a5bda278ecb407542

game/src/main/kotlin/content/skill/summoning/familiar/RavenousLocust.kt
GregHib: It comes in as: 
```
DEBUG [InstructionTask] Greg Tile(3117, 3518, 0) - InteractNPC(npcIndex=10171, option=2)
DEBUG [Player] Greg Tile(3117, 3518, 0) - Attack:guard_edgeville_7 target=NPC(id=guard_edgeville_7, index=10171, tile=Tile(3114, 3512, 0))
```
So I think you're correct, maybe it needs to hijack the Attack option

PR #1064, opened 2026-07-08
Add random events (framework + events)

HarleyGilpin:
## Overview

Implements the anti-macro **Random Events** D&D (ref GregHib/void-private#223). This is the main PR for the feature: it lands the shared framework and then accumulates each of the 18 current (2011) random events as they're built, verified against the wiki (`oldid=3667851`).

## Framework

- **Lifecycle** (`content/activity/event/random/RandomEvents`): weighted `pick()` over the `random_events` table, `start`/`complete`/`fail`, persisted per-player cooldown, and `noteAndTeleport` (ignore penalty — notes the inventory then exiles the player so an event can't be macroed through).
- **Trigger** (`RandomEventTrigger`): rolls per XP drop (1-in-N), gated by an `eligible()` check (no bots, combat, instances, dialogue, or `no_random_events` areas), plus a `::randomevent [event]` admin command.
- **Entry** (`RandomEventKidnap`): relog resume, teleport-out blocking, the `kidnap(tile)` helper, and the `mysteriousOldMan()` intro. Two supported patterns:
  - *Kidnap-to-instance/room* — the player is teleported to an event area (Maze, Freaky Forester).
  - *In-place nagging* (`InPlaceRandomEvent`) — the event NPC spawns beside the player, follows and nags, and applies the ignore penalty on timeout (Certer, Sandwich Lady, Evil Twin...).

## Extras
- Add Iffie's costume store 
- Add random event gift rewards interface
- Add XP lamps

## Media / Screenshots
image

image

## Events

- [x] **Maze** — GregHib/void-private#257
- [x] **Freaky Forester** — GregHib/void-private#255
- [x] **Certer** — GregHib/void-private#253
- [x] **Drill Demon** — GregHib/void-private#254
- [x] **Kiss the Frog** — GregHib/void-private#256
- [x] **Mime** — GregHib/void-private#258
- [x] Cap'n Arnav's Chest — GregHib/void-private#259
- [x] **Quiz Master** — GregHib/void-private#260
- [x] **Sandwich Lady** — GregHib/void-private#261
- [x] **Evil Bob** — GregHib/void-private#262
- [x] **Surprise Exam** — GregHib/void-private#263
- [x] Bee-Keeper — GregHib/void-private#264
- [x] Evil Twin — GregHib/void-private#265
- [x] Pinball — GregHib/void-private#266
- [x] Gravedigger — GregHib/void-private#267
- [x] **Pillory** — GregHib/void-private#268
[...]

PR #1070, opened 2026-07-11
Random events/bug fixes and tweaks

HarleyGilpin:
Follow-up fixes and polish for the random events added in #1064.

## Reward duplication fixes

Several reward handlers granted items and then suspended on a dialogue, only clearing their gating state after the player clicked continue. Re-triggering the entry point (talking again / re-opening the gate) cancelled the suspended coroutine before the state was cleared, allowing rewards to be claimed repeatedly. State is now cleared before the first suspension in each case, with regression tests that fail against the old code.

- [Sandwich Lady: talking to her again instead of clicking continue handed out free food and a gift each time](https://github.com/GregHib/void/commit/dd412deaa4a8406144ae2dd099f6a1eebb377790). The knockout fade path got the same hardening.
- [Explorer Jack: asking to claim Task System rewards again mid-dialogue paid out repeatedly](https://github.com/GregHib/void/commit/e369d8d759b5864c6c90b3fc390fa1907b7d600d). This commit also fixes coin payouts being priced by the final progress total instead of each task's own index (batched claims overpaid), and matches the official dialogue: an apology when nothing is owed, and rewards granted individually so anything that doesn't fit stays owed ("I've held on to the rest") instead of failing the whole claim.
- [Prison Pete: re-opening the exit gate during the farewell dialogue re-ran the escape with the keys still counted, granting unlimited presents](https://github.com/GregHib/void/commit/daf2457223da1fe20a17a8c9de13c6ed2b2bd754).

All other random events were audited for the same pattern and are safe - they either clear state before suspending or grant with no suspension in the window. A guard test documents that the Maze shrine is not exploitable (the teleport delay blocks input).

## Cutscene interruption fixes

Camera cutscenes pan the camera, `delay()`, then reset it. If the delay coroutine is abandoned before the reset, the camera stays frozen in place. Input handlers drop clicks while the player is delayed, but several paths bypassed that, so interacting during a cutscene could abandon it (frozen camera) or walk the player off.

- [Item-on-item clicks nulled the player's suspension unconditionally, abandoning the running cutscene](https://github.com/GregHib/void/commit/ee81c36b199e5bd773f39413ceb01c616744ef52).
- [The item-on-object / NPC / player / floor-item handlers had no delay guard either, so using an item on a target mid-cutscene froze the camera the same way](https://github.com/GregHib/void/commit/c9e52a73948eaae71ef1d6dd6e15e70a9f1c8f35). All four now behave like the plain option handlers.
- [Evil Bob's servant hint waited for a continue click before the pan's delay was set, leaving a window to click the fishing spot and walk off (or freeze the camera)](https://github.com/GregHib/void/commit/36f91933d46c985c88d39d5b0cbd4773742626e9). The hint line is now non-continuable so the pan locks the player on the same tick it appears; talking again after the spot has been shown just gives a reminder.

## Bank deposit boxes

- [ScapeRune's bank deposit box (object 32930) had no config entry, so its string id never matched the `bank_deposit_box*` handler and the Deposit option did nothing](https://github.com/GregHib/void/commit/d6d3a857c4eb5d4800fe12963358ea048efaba7b).
- [Same fix applied to the unnamed deposit boxes in Cap'n Arnav's and Freaky Forester's event areas](https://github.com/GregHib/void/commit/6dd859f9c6b0edc1ea725cdd536ec1b7d355da14).

## Evil Bob

- [Re-show the fishing spot hint after a relog](https://github.com/GregHib/void/commit/01e9f8686f16132bff8be950f4b82747af6f0f35)
- [showSpot camera adjustments](https://github.com/GregHib/void/commit/3da19896f5288380e32f7027f4085e4d8e115060)
- [Unstick the relog test](https://github.com/GregHib/void/commit/e0172f7b890dc760f581c5d8720f8bfa5345a4d0)
- [Fish-like things can't be eaten or banked](https://github.com/GregHib/void/commit/a20054388c7f814e7da82c1ec242cd1eac90abb5): eating one now gives the disgusted refusal, and depositing one in the island's box destroys it instead of banking it (where it would otherwise survive past the event). Deposit-destroy is gated on a new `unbankable = 2` value.

## Dialogue and polish

- [Use talkWith(npc) instead of npc ids in random event dialogues](https://github.com/GregHib/void/commit/a2c728045734be464cbd0fb8c4d8d94eb3b8321c), with [one-liner dialogues kept on the npc id form](https://github.com/GregHib/void/commit/0a7ef72031b0d1741cd14507fb2f3c0154391a44)
- [Correct teleport-out graphic when leaving random events](https://github.com/GregHib/void/commit/86f445095c19472b1451edbbd3ce722aedc2a7fa)


game/src/main/kotlin/content/activity/event/random/capn_arnav/CapnArnav.kt
```diff
@@ -72,11 +74,19 @@ class CapnArnav : Script {
                 return@objectOperate
             }
             if (!get("arnav_solved", false)) {
-                npc<Angry>("capn_arnav", "Ye're not goin' anywhere till ye've opened me chest, matey!")
+                talkWith(NPCs.find(tile.regionLevel, "capn_arnav"))
+                npc<Angry>("Ye're not goin' anywhere till ye've opened me chest, matey!")
```
GregHib: One's like these are fine because they're one liners, it's only when there's a whole conversation does it make sense to use talkWith

HarleyGilpin: Agreed - reverted the one-liners (portals, puzzle results, servant hints, drill orders) back to the npc id form and dropped the target lookups; talkWith stays only where a whole conversation follows (Evil Bob's intro): https://github.com/GregHib/void/pull/1070/commits/0a7ef72031b0d1741cd14507fb2f3c0154391a44

GregHib (approved):

PR #1072, opened 2026-07-14
Support underscore player names in console commands

HarleyGilpin:
The console input filter blocks double quotes, so the quoted-argument form (`::give "player name" ...`) is untypeable in game. Several player-targeting commands did their own exact-match name lookups, making players with spaces in their names unreachable: `give`, `tele_to`, `tele_to_me`, `rights`, `mute`/`perm_mute`/`unmute` and `ban`/`perm_ban`/`unban`.

All of them now share the underscore-for-space lookup already used by the other commands (`::tele_to eddies_fan` finds "eddies fan") - case-insensitive, prefix matching included, with a new full-name tier so a prefix can't shadow the intended target of a moderation command (`eddies_fan` can't accidentally match "eddies fanatic"). Ban's offline account lookups translate underscores the same way.

Covered by `GiveCommandTest` and `SpacedNameCommandsTest` (tele_to, mute, and the exact-beats-prefix case).

### DIFF
--- a/data/area/kandarin/feldip_hills/feldip_hills.npcs.toml
+++ b/data/area/kandarin/feldip_hills/feldip_hills.npcs.toml
@@ -109,6 +109,7 @@ immune_poison = true
 [jungle_strykewyrm]
 id = 9467
 combat_def = "jungle_strykewyrm"
+slayer_level = 73
 examine = "Who said worms were small and harmless?"
 
 [swamp_toad_feldip_hills]
--- a/data/skill/slayer/slayer.tables.toml
+++ b/data/skill/slayer/slayer.tables.toml
@@ -219,6 +219,9 @@ npc = "jelly"
 [.jungle_horrors]
 npc = "jungle_horror"
 
+[.jungle_strykewyrm]
+npc = "jungle_strykewyrm"
+
 [.kalphites]
 npc = "kalphite_worker"
 combat_level = 15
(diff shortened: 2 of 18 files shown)


GregHib (changes requested):
Couldn't Players.find just be updated to replace underscores with spaces in the search? I don't think usernames can have underscores so shouldn't cause any issues

HarleyGilpin:
Good call - done in 56ef102351a8f856dee9ac5ae9cb399569bd4f74. `Players.find` now replaces underscores with spaces before matching, the extra `search` tiers are gone, and the ban commands are back on plain `Players.find`. The console-command `find(player, name)` overload applies the same translation before its case-insensitive/prefix match, so the exact-name-beats-prefix behaviour is unchanged (still covered by `SpacedNameCommandsTest`).

GregHib (changes requested):
Let's also revert the changes where remove messages() in commands as they're useful to admin/mods using them rather than just failing silently

HarleyGilpin:
Reverted in a186416e9bb70d3297de4357893d95987b870029 - the not-found messages are back in each command (give, rights, tele_to/tele_to_me, mute/perm_mute/unmute), with the lookup going through `Players.find` which handles the underscore translation. Also made `Players.find` case-insensitive so `tele_to`/`rights` keep the case-insensitive matching they had before.

For what it's worth, the `Players.find(player, name)` overload those commands were calling does send an "Unable to find player '$name' online." console message itself rather than failing silently - but the explicit per-command messages are clearer either way.

GregHib (approved):

PR #1075, opened 2026-07-15
fix: gravestone loot despawning in seconds instead of minutes

### LINKED ISSUE #1074: Player items despawn in 39 seconds
When a player dies, their dropped items become public loot almost instantly (within seconds). This is because Gravestone.spawn() returns minutes but the value is used directly as ticks.

**To Reproduce**
1. Die to a random NPC (I died to a guard west of Lumbridge)
2. Items become public in 3 seconds.
3. Items vanish after 39 seconds. (Specifically, the memorial plaque)

**Expected behavior**
Public after 5 minutes, vanish after 15.

**Desktop (please complete the following information):**
- OS: Windows 11
- Version 26200.8737

HarleyGilpin:
## Problem

When a player dies, their dropped items go public within seconds and vanish after ~39s instead of lasting the gravestone's lifetime.

`Gravestone.spawn()` returned the gravestone lifetime in **minutes**, but `PlayerDeath.dropItems` uses that value as a **tick** count for the dropped items' reveal and disappear timers. A 5-minute memorial plaque became 5 ticks, so loot revealed after ~3s and disappeared after ~39s (5 + 60 ticks).

Closes #1074

## Fix

Return the lifetime in ticks (`TimeUnit.SECONDS.toTicks(seconds)`), matching the repair/bless path in `Gravestones.updateItems`, which already sets the floor items' timers in ticks. A memorial plaque now keeps loot private for its full 5 minutes, and the longer gravestones (up to the royal dwarven's 15) scale correctly.

## Test

`GravestonesTest` now kills a player and asserts the dropped coins get a reveal timer near 500 ticks (5 minutes), not 5. The test fails against the pre-fix code.

### DIFF
--- a/game/src/main/kotlin/content/area/misthalin/lumbridge/church/Gravestone.kt
+++ b/game/src/main/kotlin/content/area/misthalin/lumbridge/church/Gravestone.kt
@@ -9,6 +9,7 @@ import world.gregs.voidps.engine.entity.character.player.Player
 import world.gregs.voidps.engine.entity.character.player.male
 import world.gregs.voidps.engine.entity.character.player.name
 import world.gregs.voidps.engine.timer.epochSeconds
+import world.gregs.voidps.engine.timer.toTicks
 import world.gregs.voidps.type.Tile
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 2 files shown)


GregHib (approved):

HarleyGilpin:
@GregHib - I just ran spotless apply to fix failed testing.

Sulrwin:
Yay 😁

PR #1095, opened 2026-07-17
fix: player name invisible in chatbox upon login

DesecratedTree:
Root cause: The chatbox CS2 script reads the player's name from the local player's appearance data. But the appearance data is only sent during the game tick via PlayerUpdateTask. The chatbox opens during spawn() before the first tick runs, so the client has no name and falls back to [quickchaticon].

Fix in AccountManager.kt:89: Added PlayerUpdateTask().run(player) after loadCallback (which adds the player to the viewport tracking set) but before player.open(gameFrame). This manually sends the player update (including appearance data with displayName) to the client before the chatbox opens.

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/data/AccountManager.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/data/AccountManager.kt
@@ -4,6 +4,7 @@ import world.gregs.voidps.engine.client.message
 import world.gregs.voidps.engine.client.ui.InterfaceOptions
 import world.gregs.voidps.engine.client.ui.Interfaces
 import world.gregs.voidps.engine.client.ui.open
+import world.gregs.voidps.engine.client.update.player.PlayerUpdateTask
 import world.gregs.voidps.engine.client.update.view.Viewport
 import world.gregs.voidps.engine.client.variable.PlayerVariables
 import world.gregs.voidps.engine.data.definition.*
@@ -15,6 +16,7 @@ import world.gregs.voidps.engine.entity.character.player.Player
 import world.gregs.voidps.engine.entity.character.player.Players
 import world.gregs.voidps.engine.entity.character.player.appearance
 import world.gregs.voidps.engine.entity.character.player.equip.AppearanceOverrides
+import world.gregs.voidps.engine.entity.character.player.flagAppearance
 import world.gregs.voidps.engine.entity.character.player.name
 import world.gregs.voidps.engine.entity.character.player.skill.level.PlayerLevels
 import world.gregs.voidps.engine.event.AuditLog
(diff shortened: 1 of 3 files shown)


skalfate:
What an obscure bug. Interesting how you came to that one.

DesecratedTree:
> What an obscure bug. Interesting how you came to that one.

Just a minor annoyance I've been chasing for a while. Noticed it a few times while playing live server.

engine/src/main/kotlin/world/gregs/voidps/engine/data/AccountManager.kt
```diff
@@ -60,6 +62,7 @@ class AccountManager(
         player.body.link(player.equipment, overrides)
         player.body.updateAll()
         player.appearance.displayName = player.name
+        player.flagAppearance()
```
GregHib: This isn't actually necessary as AppearanceEncoder has `initial = true`
```suggestion
```

engine/src/main/kotlin/world/gregs/voidps/engine/data/AccountManager.kt
```diff
@@ -15,6 +16,7 @@ import world.gregs.voidps.engine.entity.character.player.Player
 import world.gregs.voidps.engine.entity.character.player.Players
 import world.gregs.voidps.engine.entity.character.player.appearance
 import world.gregs.voidps.engine.entity.character.player.equip.AppearanceOverrides
+import world.gregs.voidps.engine.entity.character.player.flagAppearance
```
GregHib:
```suggestion
```

engine/src/main/kotlin/world/gregs/voidps/engine/data/AccountManager.kt
```diff
@@ -83,9 +86,10 @@ class AccountManager(
             logout(player, false)
         }
         loadCallback.invoke(player)
+        Spawn.player(player)
+        PlayerUpdateTask().run(player)
         player.open(player.interfaces.gameFrame)
         player.variables.sendAll()
```
GregHib:
```suggestion
        player.open(player.interfaces.gameFrame)
        player.variables.sendAll()
        Spawn.player(player)
        // Reload game-frame to reveal display name
        player.strongQueue("login", 2) {
            player.open(player.interfaces.gameFrame)
        }
```

GregHib (changes requested):

engine/src/main/kotlin/world/gregs/voidps/engine/data/AccountManager.kt
```diff
@@ -4,6 +4,7 @@ import world.gregs.voidps.engine.client.message
 import world.gregs.voidps.engine.client.ui.InterfaceOptions
 import world.gregs.voidps.engine.client.ui.Interfaces
 import world.gregs.voidps.engine.client.ui.open
+import world.gregs.voidps.engine.client.update.player.PlayerUpdateTask
```
GregHib:
```suggestion
```

GregHib (approved):

GregHib:
> Just a minor annoyance I've been chasing for a while.

Thanks, yeah this had been annoying me for the longest time too but never knew what the cause was, I've tweaked your changes as it could be simplified further

PR #1096, opened 2026-07-17
fix: map blackscreen after teleport from differing height levels

DesecratedTree:
Root cause identified: Client's method3157 has a guard condition that states if zoneX == storedZoneX && zoneY == storedZoneY && !forceRefresh, the body is skipped, leaving scene data cleared by method3120 but never repopulated → black map.

Fix applied: Added Z-level check to inViewOfZone in RegionLoading.kt:99 which previously only checked X/Y, so teleports to the same zone coordinates but different plane (e.g., surface↔dungeon) would NOT trigger a region reload, causing the player to have a black map. 
image

### DIFF
--- a/game/src/main/kotlin/content/entity/world/RegionLoading.kt
+++ b/game/src/main/kotlin/content/entity/world/RegionLoading.kt
@@ -96,7 +96,7 @@ class RegionLoading(val dynamicZones: DynamicZones) : Script {
     /**
      * Check if we're within 4 (default) zones of the last loaded zone.
      */
-    fun inViewOfZone(player: Player, zone: Zone, radius: Int): Boolean = Distance.within(player.tile.zone.x, player.tile.zone.y, zone.x, zone.y, radius)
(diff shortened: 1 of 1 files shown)


skalfate:
This is lovely. I am amazed.

GregHib:
Hmm I've never encountered this before, would it trigger using ladders to change levels?

DesecratedTree:
> Hmm I've never encountered this before, would it trigger using ladders to change levels?

Good question, I've never tested ladders. I'm assuming its just teleports since the same map square is already loaded on level changes (ladders). Test on live by going up to lumbridge castle and teleporting to camelot or another area. You will see the bug.

GregHib:
I'm unable to replicate this issue, can you test with the normal client rather than the runelite one? Need to rule that out.

Also it doesn't make sense to me how this fix helps. If you're teleporting from lumbridge castle to camelot the region coordinates will change and `Distance.within` will return false triggering a region load. This change won't effect that scenario.

The only time a player changes level but not region is when going up/down stairs

skalfate:
I can't replicate it either on the latest client in standalone form.

GregHib:
Closing, if it's still occurring and you find out how to replicate it please open another ticket

Issue #1088, opened 2026-07-16
Rune Essence Mine - 2x Essence Bug


skalfate:
**Describe the bug**
At the Rune Essence Mine, when a Player is rewarded a Essence, at level 25/35 for Pure Essence, the player gets a Pure essence and a Rune Essence for one successful reward.

**To Reproduce**
1. Go to Rune Essence Mine and Mine essence

**Expected behavior**
Only 1 Pure Essence should be reward for qualified leveled player.

**Screenshots**

<img width="1920" height="1048" alt="Image" src="https://github.com/user-attachments/assets/e0a5b31a-6669-4e62-8332-00c87923836a" />

<img width="1920" height="1048" alt="Image" src="https://github.com/user-attachments/assets/66fcb00a-5f9c-4a06-ad16-8296bfc7b488" />

**Desktop (please complete the following information):**
- OS: [Windows 7]
- Version [e.g. 1.0.0]

**Additional context**
None

PR #1097, opened 2026-07-17
Fixed mining rune essence giving both a rune essence and pure essence…

skalfate:
Fixed mining rune essence giving both a rune essence and pure essence together per success at level 30 and after. Should fix issue #1088

### DIFF
--- a/game/src/main/kotlin/content/skill/mining/Mining.kt
+++ b/game/src/main/kotlin/content/skill/mining/Mining.kt
@@ -105,7 +105,7 @@ class Mining : Script {
                     }
                 }
                 for (item in ores) {
-                    if (item == "rune_essence" && has(Skill.Mining, 30)) {
+                    if (item == "rune_essence" && has(Skill.Mining, 30) && World.members) {
                         continue
                     }
                     if (item == "pure_essence" && !World.members ) {
--- a/game/src/main/kotlin/content/skill/mining/Mining.kt
+++ b/game/src/main/kotlin/content/skill/mining/Mining.kt
@@ -105,7 +105,10 @@ class Mining : Script {
                     }
                 }
                 for (item in ores) {
-                    if (item == "pure_essence" && !World.members) {
+                    if (item == "rune_essence" && has(Skill.Mining, 30)) {
+                        continue
+                    }
+                    if (item == "pure_essence" && !World.members ) {
                         continue
                     }
                     val ore = Rows.getOrNull("ores.$item") ?: continue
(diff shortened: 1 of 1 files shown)


game/src/main/kotlin/content/skill/mining/Mining.kt
```diff
@@ -105,7 +105,10 @@ class Mining : Script {
                     }
                 }
                 for (item in ores) {
-                    if (item == "pure_essence" && !World.members) {
+                    if (item == "rune_essence" && has(Skill.Mining, 30) && World.members) {
```
GregHib, line 108: Won't this still give both if < 30 mining?

skalfate: `Ello Greg. At first glance, the logic code may seems unclear, yes, but rest assure I tested it and it behaves as it should.

The server has a (for each) ore in (ores) function when mining. In a few circumstances, multiple ores options are faced. This applies to rune essence rocks, limestone and granite rocks, and gem rocks. Checks would likely have to implemented on Quarry and Gem rocks of Shilo Village in the future, I guess, because I reckon this bug will apply to them where they will give more than intended.

Currently in the for loop, it loops twice on every successful mine of rune essence. I shifted rune essence to the top because that executes first before pure essence, due to the order of the ores in the array. This should save an extra instruction from being processed when rune essence isn't to be given. 

If mine is successful, a rune essence will "not" be given in the following situations:

- skill level of 30+
- members world

But Rune essence will be given in free world regardless of levels, as that is authentic.

At level less than 30, it won't give both because rune essence check at the top will fail and give rune essence and then on second loop, for pure essence, line 112 does a level check for the ore and handles pure essence case.  

Ultimately the second loop runs and that will be for pure essence and in members world it will be issued and "not issued" in free to play world.

Please do reply if I a missing something.

skalfate: By the way should I add comments into this pull request? For example 

```
/**
                     * For loop runs twice in the case of mining Rune Essence because 2 ores are possible from the same rock. 
                     * 1st loop for Rune Essence. 2nd loop for Pure Essence.
                     */


                    // Checks to see if the Player has the level required for the Ore mined.

```

GregHib: Nope no comment required, thanks

GregHib (approved):

game/src/main/kotlin/content/skill/mining/Mining.kt
```diff
@@ -105,7 +105,10 @@ class Mining : Script {
...
+                        continue
+                    }
+                    if (item == "pure_essence" && !World.members ) {
```
GregHib:
```suggestion
                    if (item == "pure_essence" && !World.members) {
```

PR #1098, opened 2026-07-17
fix: dialogue / combat queue conflict

DesecratedTree:
The Bug
When a dialogue (like a level-up or a random event) opened during combat, the player and their attacker was completely frozen. Combat stopped with no movement, no attacks, no damage. The player had to manually dismiss the dialogue before combat resumed. When this would happen, damage would "pool up behind the dialogue". This is unintended behavior and reference videos alongside historical sources show that combat actually interrupts the dialogue interface. Furthermore, this adds in the chat message for leveling up and ensures the player sees the message even if the dialogue is closed by a strong queue action.

Changes
CombatMovement.kt
Removed the character.dialogue != null early return from tick().

Before: If the player had a dialogue open, CombatMovement.tick() returned immediately and the player stopped moving toward their target and stopped attacking.

After: Combat movement ticks normally regardless of dialogue state. The player continues chasing targets, pathfinding, and triggering attacks.

Combat.kt
Removed the character.dialogue != null early return from combat().

Before: Even if CombatMovement.tick() ran, the actual attack logic in Combat.combat() also bailed out when dialogue was open. No swings, no damage, no XP until the dialogue was cleared. 

After: Attack rolls, damage, and combat XP all process normally while dialogue is visible.

ActionQueue.kt
Strong actions now bypass canProcess(), and Strong actions close dialogue.

Two changes here:

process() now checks action.priority == ActionPriority.Strong || canProcess() instead of just canProcess(). This means player damage hits (strongQueue("hit")) execute immediately instead of pooling in the queue behind the dialogue interrupt check.

The existing closeMenu() call in tick() (which fires when a Strong action is pending) now also calls closeDialogue(). Getting hit is a Strong action, so it force-closes any open dialogue — matching the original RS2 behavior where combat interrupts the level-up display.

LevelUp.kt
Added a message() call alongside the level-up interface.

The congratulations text is now sent as a ChatType.Game chat message in addition to being displayed in the level-up interface. This is historically accurate to 2011 RS2, where both the chatbox message and the interface were sent on level-up.

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/combat/CombatMovement.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/combat/CombatMovement.kt
@@ -1,6 +1,5 @@
 package world.gregs.voidps.engine.entity.character.mode.combat
 
-import world.gregs.voidps.engine.client.ui.dialogue
 import world.gregs.voidps.engine.client.variable.hasClock
 import world.gregs.voidps.engine.data.config.CombatDefinition
 import world.gregs.voidps.engine.data.definition.CombatDefinitions
@@ -43,9 +42,6 @@ class CombatMovement(
     }
 
     override fun tick() {
-        if (character is Player && character.dialogue != null) {
-            return
-        }
         if (target.tile.level != character.tile.level) {
             character.mode = EmptyMode
             return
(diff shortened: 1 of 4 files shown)


skalfate:
I have tested the before and after of the Level-up, Combat.kt, and CombatMovement.kt changes and strong queue can confirm the fixes indeed fix the things mentioned. :)

GregHib:
Fixed in #1122

PR #1099, opened 2026-07-18
feat: teleport sounds

DesecratedTree:
still need to add jewelry sounds but this should work for now

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/player/Teleport.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/player/Teleport.kt
@@ -9,6 +9,7 @@ import world.gregs.voidps.engine.entity.character.player.skill.Skill
 import world.gregs.voidps.engine.entity.character.player.skill.exp.exp
 import world.gregs.voidps.engine.entity.character.sound
 import world.gregs.voidps.engine.entity.obj.GameObject
+import world.gregs.voidps.network.login.protocol.encode.playSoundEffect
 import world.gregs.voidps.engine.event.*
 import world.gregs.voidps.engine.map.collision.random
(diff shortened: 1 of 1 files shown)


skalfate:
Lovely! to see this. I am just popin in here and there. No time unfortunately.

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/player/Teleport.kt
```diff
@@ -116,6 +117,11 @@ interface Teleport {
                 player.exp(Skill.Magic, xp)
                 if (sound) {
                     player.sound("teleport_${type}")
+                    if (type == "tablet") {
```
GregHib, line 120: If you add the ids as teleport_tablet and teleport_jewellery to .sounds.toml that should work

skalfate: I am seeing the file currently, and the those ids and name that you suggested Greg already exists. Then how come it does not play the sound?

Although teleport land is et to id 201, maybe I'll test that out with changing it to 200. Nope. Going to further and debug and see what whats happening.

skalfate: Ah I found the fix. Its 👍🏻 

[teleport_modern]
id = 200

skalfate (changes requested):
Hiya Hawthorne, This is the change that is needed in the following file data/skill/magic/teleport/teleport.sounds.toml

[teleport_modern]
id = 200

Tablets and many others are already implemented.

DesecratedTree:
I will go ahead and update the PR when I get home tonight. Thank you!

GregHib:
Fixed in #1121

PR #1100, opened 2026-07-22
fix: allow inventory reordering while bank is open

### LINKED ISSUE #1060: Can't drag inventory items when bank is open
Should be able to reorder items in inv when bank is open.
I presume an inv unlock needs to be sent to enable it.

HarleyGilpin:
Closes #1060

## Problem
Dragging items to reorder the inventory while the bank (or deposit box) was open did nothing. Additionally, depositing an item when carrying multiples deposited the first matching slot instead of the one clicked (noted in the issue comments).

## Cause
Three separate gaps:

1. **Client never initiated drags on `bank_side` (763).** The 634 client requires two access mask conditions for an item drag: drag-depth bits 18-20 != 0 on the source slot and bit 21 on the drop target (`Class348_Sub44.method3304`/`method3302` in the deob). `bank_side`'s options only unlocked the deposit/examine bits. The same combination is already what makes the normal inventory tab (`Option12 = 17` + `Drag = 20` on 149) and the music playlist work.
2. **Item ids arrive crossed in `763->763` switch packets.** The client reorders its own display *before* sending, so each item id is paired with the slot it was dropped into. Validation compared the post-swap client state against the pre-swap server state and rejected the packet - leaving the client visually reordered but the server unchanged, which then desynced every later click on those slots. Interface 149 has the same crossing (its handler already swaps `fromItemId`/`toItemId` alongside the ghost-slot `-28`); 763 needs the id swap without the slot translation, since it uses real slots 0-27.
3. **Deposits ignored the clicked slot.** `moveToLimit(item.id, ...)` removes by id from the first matching slot.

## Changes
- `bank.ifaces.toml`: add `Drag-Source = 17` (bit 18, drag-depth 1) and `Drag = 20` (bit 21) to `bank_side` inventory options, making the unlocked mask 2360446 - covers both the bank and deposit box paths
- `InterfaceSwitchHandler`: swap the crossed item ids for `763->763` switches
- `BankOpen`: handle `interfaceSwap("bank_side:inventory", "bank_side:inventory")` with `inventory.swap` (exact pair, so a future drag-to-deposit onto the bank grid isn't misrouted into an inventory swap)
- `BankDeposit`: pass the clicked slot through and remove via `RemoveItem.remove(slot, id, amount)` - empties the clicked slot first, then any remainder from the earliest slots; bank-capacity clamping now happens up front via `addToLimit`
- `BankTest`: reorder while banked (including onto an empty slot, and the crossed-id packet semantics), verifies both slots are resent to the client after a swap, deposit-after-reorder, and clicked-slot deposits

Tested in-game: dragging, reordering onto empty slots, and slot-accurate deposits all work with the bank and deposit box open.

### DIFF
--- a/data/entity/player/bank/bank.ifaces.toml
+++ b/data/entity/player/bank/bank.ifaces.toml
@@ -259,7 +259,7 @@ type = "overlay_tab"
 [.inventory]
 id = 0
 inventory = "inventory"
-options = { Deposit-1 = 0, Deposit-5 = 1, Deposit-10 = 2, Deposit-* = 3, Deposit-X = 4, Deposit-All = 5, Examine = 9 }
+options = { Deposit-1 = 0, Deposit-5 = 1, Deposit-10 = 2, Deposit-* = 3, Deposit-X = 4, Deposit-All = 5, Examine = 9, Drag-Source = 17, Drag = 20 }
 
 [bank_help]
 id = 767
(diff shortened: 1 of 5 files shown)


GregHib (approved):
Lgtm

PR #1101, opened 2026-07-24
fix: stop GE claims blocking player offer saves

HarleyGilpin:
The grand_exchange_claims.offer_id foreign key referenced grand_exchange_offers.id with the default RESTRICT rule. Claims and offers are persisted by two independent, uncoordinated save paths (per-player save vs. the hourly Grand Exchange save), so a claim row can still reference an offer that the player-save path is about to delete-and-reinsert. The DELETE was blocked by the FK, failing the whole player-save transaction (xp, levels, inventory, everything) and retrying forever.

Drop the FK so the two tables' lifecycles are fully decoupled; offer_id keeps its unique index. Existing databases are migrated on startup with an idempotent ALTER ... DROP CONSTRAINT IF EXISTS, a no-op on fresh DBs.

Adds a regression test that reproduces the FK violation pre-fix.

### DIFF
--- a/database/src/main/kotlin/world/gregs/voidps/storage/Tables.kt
+++ b/database/src/main/kotlin/world/gregs/voidps/storage/Tables.kt
@@ -152,7 +152,7 @@ internal object PlayerHistoryTable : Table("player_exchange_history") {
 }
 
 internal object ClaimsTable : Table("grand_exchange_claims") {
-    val offerId = integer("offer_id").references(OffersTable.id).uniqueIndex()
+    val offerId = integer("offer_id").uniqueIndex()
     val amount = integer("amount")
     val coins = integer("coins")
 }
--- a/database/src/main/kotlin/world/gregs/voidps/storage/Tables.kt
+++ b/database/src/main/kotlin/world/gregs/voidps/storage/Tables.kt
@@ -152,6 +152,8 @@ internal object PlayerHistoryTable : Table("player_exchange_history") {
 }
 
 internal object ClaimsTable : Table("grand_exchange_claims") {
+    // Logically references OffersTable.id, but without a foreign key constraint: claims and offers
+    // are saved by independent paths, so enforcement would block the per-player offer delete/reinsert.
     val offerId = integer("offer_id").uniqueIndex()
     val amount = integer("amount")
     val coins = integer("coins")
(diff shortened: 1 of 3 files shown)


database/src/main/kotlin/world/gregs/voidps/storage/Tables.kt
```diff
@@ -152,7 +152,7 @@ internal object PlayerHistoryTable : Table("player_exchange_history") {
 }
 
 internal object ClaimsTable : Table("grand_exchange_claims") {
-    val offerId = integer("offer_id").references(OffersTable.id).uniqueIndex()
+    val offerId = integer("offer_id").uniqueIndex()
```
GregHib, line 157: If this is not a reference to the offers table then what is it? Should it just be called id?

HarleyGilpin: It is still a reference to the offers table, the column holds `OffersTable.id` (claims are keyed by `offer.id` in `GrandExchange`), so `offer_id` is the accurate name. What's removed is only the DB-level enforcement: offers are deleted and reinserted per player on save while claims persist independently, so a FK constraint blocks the offer re-save whenever an outstanding claim exists. Added a comment clarifying this in dd7dbe1b1.

GregHib: Hmm but if it's being properly deleted and re-added in order it shouldn't be an issue on reinsertion?

HarleyGilpin: Order within the transaction doesn't help here: Postgres checks a plain (non-deferred) FK at the end of each *statement*, not at commit. `saveOffers` runs `OffersTable.deleteWhere { playerId inList ids }` as its own statement, so the DELETE throws `foreign_key_violation` the moment it removes an offer row a claim references, the reinsert in the following batchUpsert statement never gets a chance to run. The StorageTest added in this PR (`Re-saving offers with an outstanding claim doesn't fail`) reproduces exactly that against Postgres.

The delete would only survive if the constraint were declared `DEFERRABLE INITIALLY DEFERRED` (checked at commit instead), but Exposed's `references()` can't emit that, so it'd need raw SQL in schema setup. The other option is dropping the delete/reinsert pattern for offers (upsert + delete-only-missing rows). Happy to do either if you'd rather keep DB-level enforcement, dropping the FK was just the smallest change.

GregHib (approved):

Issue #1087, opened 2026-07-16
Shooting Star Sometimes double spawns


skalfate:
**Describe the bug**
The shooting star sometimes spawns 2 times in an area. The extra one is always useless and doesn't do anything. This could be any of the location it spawns in.

**To Reproduce**
1. Visit Shooting Stars.

**Expected behavior**
Only 1 star should appear.

**Screenshots**

<img width="1920" height="1048" alt="Image" src="https://github.com/user-attachments/assets/938fb5ee-cfc2-43a0-b3a0-2d590af4e80b" />

**Desktop (please complete the following information):**
- Version [e.g. 1.0.0]

**Additional context**
none

GregHib:
This might be due to #1000, if you see it again can you relog and see if the second star persists?

skalfate:
Okay I'll keep that in mind thanks. I'll leave it open until I can verify just in case.

PR #1135, opened 2026-08-05
Fix zone batch update sync causing floor item ghosting

### LINKED ISSUE #1000: Fix floor item ghosting
When dynamic zones are being modified and DynamicZone.reloadCallback is happening, there's no check that an update is already in progress and two region updates can occur in the same tick, also sending two lots of batch updates (like add item) which causes this "ghosting" effect.

1. Login at 3114, 3209. 
2. Walk to wizards tower
3. Should see logs and boots in wrong location.

Image

if the door is open, you will see it flick from open to closed as you walk down the bridge.

HarleyGilpin:
Closes #1000, #1087

Root cause is a set of sync gaps in the zone batch update pipeline rather than floor items specifically, items were just the visible symptom since spawns constantly re-queue additions.

[Screencast_20260805_143328.webm](https://github.com/user-attachments/assets/10d0d7ed-76a2-494a-a2c4-a137f096bb17)

## Problems

1. **Private updates applied to the wrong zone.** All floor item packets are `private` and sent as bare single packets, which the client applies relative to its active zone, set by whatever `clearZone`/batch header happened to come last. When that pointer pointed at a neighbouring zone, the item rendered one zone (8 tiles) off in the direction of travel.
2. **Out-of-map offsets clamped instead of skipped.** `getZoneOffset` used `safeMinus`, so zones south/west of the loaded map base clamped to offset 0 and rendered 8 tiles north/east of their true tile (reproducible with the logs at 3155 in zone `389,394` after logging in at `3114,3209`).
3. **Stale batch replay.** The `encoded` map was never cleared, so mid-tick sends (region change/batch zone change) replayed the previous tick's batch, re-applying object add/removes (door flicker on the Wizards' Tower bridge) and duplicating item additions. This affects objects just the same as items, as you noted — items only made it visible.
4. **Updates lost during map load.** `updateRegion` sent zone contents immediately after the map region packet, but the client loads maps asynchronously and drops zone updates in the window before `FinishRegionLoad`, `viewport.loaded` was tracked but never gated anything, so items vanished after every region reload.
5. **Duplicate region updates per tick.** `checkReload` and `DynamicZones.reloadCallback` could both call `updateRegion` in the same tick with no in-progress check, and two send paths with different anchors (`lastBatchZone` vs `steps.previous.zone`) cleared and resent the same entered zones twice per tick.

## Changes

- Send the active zone (`UPDATE_ZONE`) before individual private updates so they always apply to the correct zone (matches partial-follows behaviour before each bare update)
- Skip zones outside the loaded map instead of clamping their offsets
- Clear `encoded` alongside `batches` each tick and skip encoding empty batches (also fixes unbounded growth)
- Single tick-end send path anchored on `lastBatchZone`; removed the `steps.previous`-anchored duplicate
- Full zone content resend moved to the `FinishRegionLoad` handler, with sends gated on `viewport.loaded` so nothing is fired into the load window
- `DynamicZones.version` stamped on the viewport per region update so `crossedDynamicBoarder` no longer force-reloads players already updated for the current dynamic state

### DIFF
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/client/update/CharacterUpdateTask.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/client/update/CharacterUpdateTask.kt
@@ -17,7 +17,7 @@ class CharacterUpdateTask(
     override fun predicate(character: Player): Boolean = character.networked
 
     override fun run(character: Player) {
-        ZoneBatchUpdates.run(character)
+        ZoneBatchUpdates.send(character)
         playerUpdating.run(character)
         npcUpdating.run(character)
         character.viewport!!.shift()
--- a/engine/src/main/kotlin/world/gregs/voidps/engine/client/update/batch/ZoneBatchUpdates.kt
+++ b/engine/src/main/kotlin/world/gregs/voidps/engine/client/update/batch/ZoneBatchUpdates.kt
@@ -11,6 +11,7 @@ import world.gregs.voidps.network.login.protocol.encode.clearZone
 import world.gregs.voidps.network.login.protocol.encode.encodeBatch
 import world.gregs.voidps.network.login.protocol.encode.send
 import world.gregs.voidps.network.login.protocol.encode.sendBatch
+import world.gregs.voidps.network.login.protocol.encode.updateZone
 import world.gregs.voidps.network.login.protocol.encode.zone.ZoneUpdate
 import world.gregs.voidps.type.Zone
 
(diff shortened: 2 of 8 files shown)


GregHib (changes requested):
Found a bug:
1. Login
2. Run > 8 tiles away (i.e. west)
3. Drop an item

Expected:
- Item spawns on floor

Actual:
- Item is removed from inventory but not seen on floor by owner until turned visible to all

HarleyGilpin:
Found and fixed in a99c65647. the `UPDATE_ZONE` (41) encoder had x/y swapped. It was previously unused so it had never been exercised: the client reads packet 41 as y (byte inverse), level (byte add), x (byte add), but the encoder wrote x first. Dropping in place worked because the player sits at the symmetric centre offset (6,6) where the swap is invisible; running 8+ tiles west makes x=5/y=6, so the active zone landed one zone off and the private item add rendered in the wrong zone until the public reveal took the (correct) batch path.

Verified against the deob: opcode 33 and 48 read orders match the existing `clearZone`/`sendBatch` encoders.

GregHib (approved):
Tested the original scenario along with common situations with multiple players and instances. LGTM!

PR #1150, opened 2026-08-09
Fix slayer task assignment and master recommendations

HarleyGilpin:
Talking to Mazchna in Canifis and asking for a task throws:

```
ERROR [Script$Companion] Error in script handler.
java.lang.IllegalStateException: No task found for achtryn
    at content.skill.slayer.SlayerKt.assignTask(Slayer.kt:69)
    at content.skill.slayer.master.SlayerMasterKt.roll(SlayerMaster.kt:131)
    at content.skill.slayer.master.SlayerMasterKt.assignTaskDialogue(SlayerMaster.kt:109)
```

Fixing that turned up three more problems in the same dialogue, all in how a master decides who to send a strong player to.

## The crash: transforming masters

Canifis's master isn't spawned as Mazchna - it's npc `8481`, a transforming npc keyed on varbit `5525` (While Guthix Sleeps) which switches between `8464` Mazchna and `8465` Achtryn, her stand-in while she's away. Shilo Village is the same shape: `8483` transforms to `8466` Duradel / `8467` Lapalok.

The transform bases were named after the stand-in (`achtryn` = 8481, `lapalok_shilo_village` = 8483) with the real npcs pushed onto suffixed ids (`achtryn_3`, `lapalok_shilo_village_3`), and the spawns pointed at the bases.

`PlayerOnNPCInteract` looks handlers up under the **transformed** definition's id, which is why "Talk-to" and "Get-task" fired at all - only 8464/8465 carry those options, the base has just "Examine". `SlayerMaster` then passed `target.id`, the **spawned** id, into the roll. There is no `achtryn_slayer_tasks` table, so `rollTask` returned null and `assignTask` threw.

String ids now match the cache names:

| id | before | after |
|---|---|---|
| 8481 | `achtryn` | `mazchna_canifis` |
| 8465 | `achtryn_3` | `achtryn` |
| 8483 | `lapalok_shilo_village` | `duradel_shilo_village` |
| 8467 | `lapalok_shilo_village_3` | `lapalok` |

Spawns and `wander_ranges` keys follow the renames.

`SlayerMaster` resolves the master from `npc.def(player).stringId`, falling back to `npc.id` when that isn't a known master - mirroring interaction dispatch. The fallback matters for Sumona, who is also a transform base (`7779` -> `7780` `sumona_2` once Smoking Kills is done) and would otherwise break the same way.

`baseSlayerMaster()` maps a stand-in to the master whose task list they hand out (spria/turael, achtryn/mazchna, lapalok/duradel), and the stand-ins are registered for Talk-to, Get-task, Trade and Rewards. `NPCContact` drops its Lapalok special case for the same helper, and its spawn lookup points at the renamed transform bases.

## Who a master sends you to

Duradel and Kuradal fell through to a default of combat 128 paired with "Duradel in Shilo Village", so Duradel sent players to himself and Kuradal - stronger than Duradel - sent them to Duradel. The master each one pointed at was also off by a rung: Mazchna sent players past Vannaka to Chaeldar, Vannaka past Chaeldar to Sumona, Chaeldar past Sumona to Duradel. And the thresholds (50, 75, 90, 100, 120) were invented - they don't match what the rewards interface tells players:

| master | location | requirement |
|---|---|---|
| Turael | Burthorpe | combat 3 |
| Mazchna | Canifis | combat 20 |
| Vannaka | Edgeville Dungeon | combat 40 |
| Chaeldar | Zanaris | combat 70 |
| Sumona | Pollnivneach | 35 Slayer, after Smoking Kills |
| Duradel | Shilo Village | 50 Slayer, combat 100 |
| Kuradal | Ancient Cavern | 75 Slayer, combat 110 |

Those requirements are now one ordered list, and a master recommends the **strongest** master in it the player qualifies for - checking combat level, Slayer level and quest - rather than the next one up. A maxed player asking Turael for work is sent straight to Kuradal instead of being walked through four masters to get there. Chaeldar points at Duradel while Smoking Kills is outstanding rather than at a Sumona who won't hand out tasks, and Kuradal points at nobody.

Worth a second opinion: Turael now starts recommending someone else above combat 20, the interface's requirement for Mazchna, where the invented threshold stayed quiet until 50.

## Testing

`SlayerMasterTest` takes a task from both transforming masters, takes one from Kuradal through the full dialogue, and covers the recommendation: a maxed player sent to Kuradal from every rung, the best-qualified pick when Smoking Kills is outstanding and Duradel is out of reach, Sumona promoted over Chaeldar once the quest is done, and a fresh player left with Turael.

`NPCContactTest`'s Duradel case is renamed - its old name asserted the mislabeled premise. Full suite passes.


GregHib (approved):

game/src/test/kotlin/content/skill/slayer/SlayerMasterTest.kt
```diff
@@ -41,38 +41,65 @@ class SlayerMasterTest : WorldTest() {
...
         for (skill in Skill.entries) {
-            player.levels.set(skill, 99)
+            player.setLevel(skill, 99)
         }
         return player
     }
+
+    private fun Player.setLevel(skill: Skill, level: Int) {
+        experience.set(skill, Level.experience(level))
+        levels.clear(skill)
```
GregHib: This won't set level to 99

HarleyGilpin: Swapped it for the pattern the other tests use: set the level, then derive experience from it with the skill-aware `Level.experience(skill, level)` overload. `maxedPlayer` now uses `skill.maximum()` so Constitution ends up at 990 rather than 99.

cea5662c6

PR #1182, opened 2026-08-18
Wyson mole part exchange, giant mole burrow crash and light source fixes

HarleyGilpin:
Three related changes that came out of working on the Giant Mole and Wyson.

## Wyson the gardener

Completes his conversation tree from the [wiki transcript](https://runescape.wiki/w/Transcript:Wyson_the_gardener), replacing the `// TODO add selling mole parts`:

- The unusual flower bed branch, including the "Why are you even collecting mole claws?" detour
- Giant Mole combat advice
- "Tell me your price." , bulk woad leaves at 25gp each
- The mole part exchange

Mole claws and skins trade for cheap seed nests, noses for seed nests, matching the item split in game. Completing the Falador hard tasks adds a white lily seed per skin.

Parts are swapped one at a time in their own transaction, so a full backpack stops the trade rather than rolling all of it back, a skin plus its seed needs two slots for the one it frees.

Two deliberate deviations from the wiki:

- Nests are given unnoted. No noted bird nest exists in the cache (`noteId = -1` on every nest), so there is nothing to hand out.
- The 5 and 10 coin refusals end the conversation instead of looping back to the price menu. That matches the transcript, and the menu is already at the five option cap so "changed my mind" has nowhere to go.

## Giant Mole burrow crash

`gianMoleSpawns.random(mole)` was dereferenced with `!!`. `Area.random` gives up after 100 attempts to fit a size 3 npc and returns null, so this threw, and it threw *between* the burrow down and burrow up animations, leaving the mole rendered underground permanently. It now surfaces in place and logs a warning.

## Falador task varbits

`i_heard_you_like_mudskips` and `these_arent_the_coins_youre_looking_for` had their varbit ids swapped, so the hard task check read the wrong bit.

Task structs carry a task index whose 115..151 range maps contiguously onto varbits 5691..5727 ordered by index. Every other entry in the file matches that mapping; those two were an exact swap of each other.

Also renames `falador_easy.varbits.toml` to `falador_tasks.varbits.toml`, since it holds all four difficulty sets and every other area already uses the `_tasks` name.

## Extinguishing light sources

`Light.extinguish` built its column path as `row.item(".unlit")`. `RowDefinition.item` already prepends the row id, so the path became `extinguish.<item>..unlit`; `Tables.get` splits on `.` and destructures the first three parts, leaving the column empty and failing with `Column  not found`.

This threw for *any* lit candle or torch being snuffed out, so no light source could be extinguished anywhere in the game. Caught in production via the Giant Mole's dirt attack.

## Dirt-on-screen resilience

That crash surfaced a second problem in `handleDirtOnScreen`. The overlay was opened for every nearby player *before* the timer that closes it was queued, so a throw part way through the open loop skipped the queue entirely and left everyone in range stuck behind a dirt screen until they logged out. The timer is now registered first, and each player is guarded individually.

The close loop is guarded too. World queue blocks run unguarded inside a game loop stage, and `GameLoop` catches outside its `while` loop, so a throw while closing would have ended the game loop rather than just one overlay.

## Testing

`./gradlew test` passes. New coverage:

- `FiremakingTest` - candles and torches swap to their unlit form, lanterns stay lit
- `GiantMoleTest` -nearby players are blinded, their lit light sources are snuffed, and the overlay clears itself

### DIFF
--- a/game/src/main/kotlin/content/area/asgarnia/falador/WysonTheGardener.kt
+++ b/game/src/main/kotlin/content/area/asgarnia/falador/WysonTheGardener.kt
@@ -7,11 +7,11 @@ import content.entity.player.dialogue.type.intEntry
 import content.entity.player.dialogue.type.item
 import content.entity.player.dialogue.type.npc
 import content.entity.player.dialogue.type.player
+import content.entity.player.inv.item.addOrDrop
 import world.gregs.voidps.engine.Script
 import world.gregs.voidps.engine.client.message
 import world.gregs.voidps.engine.entity.World
 import world.gregs.voidps.engine.entity.character.player.Player
-import world.gregs.voidps.engine.entity.item.floor.FloorItems
 import world.gregs.voidps.engine.inv.*
 import world.gregs.voidps.engine.inv.transact.operation.AddItem.add
 import world.gregs.voidps.engine.inv.transact.operation.RemoveItem.remove
@@ -124,9 +124,7 @@ class WysonTheGardener : Script {
     suspend fun Player.buyWoadLeaf() {
         npc<Idle>("Mmmm... okay, that sounds fair.")
         if (inventory.remove("coins", 15)) {
-            if (!inventory.add("woad_leaf")) {
-                FloorItems.add(tile, "woad_leaf", disappearTicks = 300, owner = this)
-            }
+            addOrDrop("woad_leaf")
             player<Happy>("Thanks.")
             item("woad_leaf", "You buy a woad leaf from Wyson.")
             npc<Idle>("I'll be around if you have any more gardening needs.")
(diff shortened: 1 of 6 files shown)


game/src/main/kotlin/content/area/asgarnia/falador/WysonTheGardener.kt
```diff
@@ -1,99 +1,285 @@
...
+    suspend fun Player.tooLittle() {
+        npc<Angry>("No no, that's far too little. Woad leaves are hard to get. I used to have plenty but someone kept stealing them off me.")
+    }
+
     suspend fun Player.buyWoadLeaf() {
         npc<Idle>("Mmmm... okay, that sounds fair.")
         if (inventory.remove("coins", 15)) {
             if (!inventory.add("woad_leaf")) {
                 FloorItems.add(tile, "woad_leaf", disappearTicks = 300, owner = this)
             }
```
GregHib: FYI there's a addOrDrop() helper that can be used for this when a transaction isn't needed

GregHib (approved):

game/src/main/kotlin/content/area/asgarnia/falador/WysonTheGardener.kt
HarleyGilpin: Swapped both woad leaf purchases to `addOrDrop()`. 536b7e8d8ac40cde4374dbf5827d4eed7cc6addc

PR #1186, opened 2026-08-18
Dragonfire shield charging, discharge and creation

HarleyGilpin:
Implements the dragonfire shield, which had item definitions but no behaviour.

Referenced against the [2011 wiki revision](https://runescape.wiki/w/Dragonfire_shield?oldid=4845107)

## Shield

- **Charging** - every dragonfire and skeletal wyvern icy breath the shield absorbs adds a charge, up to 50. An uncharged shield becomes charged on the first one. Each charge adds one to its melee, ranged and summoning defences.
- **Activate** - the worn option spends a charge to breathe dragonfire back at the current target for up to 290, on a two minute cooldown, at up to twelve squares and over obstacles, granting Magic, Defence and Constitution experience. Dragons are immune.
- **Inspect** reports the charge count, **Empty** releases the lot and reverts to the tradeable uncharged shield.
- Charges are lost with the shield on death, worked out after items kept on death so a protected shield stays charged.

## Creation

- Smith a draconic visage onto an anti-dragon shield at an anvil, needing 90 Smithing and a hammer, for 2000 Smithing experience.
- Or pay Oziach 1,250,000 coins to forge it. He had no dialogue at all before this, only the shop his definition points at.

Both produce the uncharged shield.

## Supporting fixes

Each of these was reached, or made reachable, by the shield.

- **`Item.amount` counted charges as a quantity.** The guard only recognised item-value charges when the definition's *starting* charge count was above one, which it never is for something that starts empty. A twenty charge shield therefore looked like twenty items: items kept on death protected three of them and the rest dropped as a stack. Capacity now decides, which also covers satchels, tea flasks and enchanted headgear.
- **Dragonfire reduction in `Equipment` read the wrong entity.** It sat in a `when (source)` and used the attacker's shield and antifire, messaging the attacker. Unreachable while only npcs breathed dragonfire, but firing the shield woke it up and cut the wielder's own damage by 80% while telling them their shield had absorbed it. `Dragonfire.maxHit` already applies the victim's shield, potion and prayer, so the duplicate is gone and its shield message moved there, where it reaches the victim.
- **Fire resistant shields did nothing against icy breath**, from the same wrong-entity bug, and `fireResistantShield` compared against `dragonfire_shield`, an alias rather than an equipped id. The cap also *replaced* the damage rather than limiting it, so a small hit was raised to it. `SkeletalWyvern` now shares the one shield check instead of keeping its own copy.
- **Player dragonfire granted Constitution experience alone**, now Magic and Defence too.
- **`Dragonfire.maxHit` returned `-1` for an immune dragon**, which would have thrown out of `random.nextInt(0, 0)` in `Damage.roll` as soon as a player could deal dragonfire.

## Data

Shield definitions moved out of `ancient_cavern.*` into `data/entity/player/equipment/dragonfire_shield.*`, since it is not ancient cavern content.

## Testing

20 tests across `DragonfireShieldTest`, `OziachTest` and `IcyBreathTest` covering charging and the cap, the defence bonus, activate with its cooldown, range and dragon immunity, inspect and empty, death conversion, both creation routes and their requirements, and the icy breath cap. Full suite green.

One test changed: `EnchantedHeadgearTest` asserted a helm's stored scroll count through `Item.amount`, now read through `charges()`.

### DIFF
--- a/game/src/main/kotlin/content/entity/player/effect/Dragonfire.kt
+++ b/game/src/main/kotlin/content/entity/player/effect/Dragonfire.kt
@@ -44,10 +44,7 @@ object Dragonfire {
         } else if (source is Player) {
             return SHIELD_MAX_HIT
         } else if (target is Player) {
-            if (Equipment.antiDragonShield(target)) {
-                target.message("Your shield absorbs most of the dragon's fiery breath!", ChatType.Filter)
-            }
-            target.message(if (success) "You're horribly burnt by the dragon fire!" else "You manage to resist some of the dragon fire!", ChatType.Filter)
+            target.message(message(target, success), ChatType.Filter)
         }
         return maxHit(
             type = type,
@@ -60,6 +57,15 @@ object Dragonfire {
                 else -> 0
             },
         )
+    }
+
+    /**
+     * Only whichever protection did the most to soften the breath is reported, they don't stack up.
+     */
+    private fun message(target: Player, success: Boolean): String = when {
+        Equipment.antiDragonShield(target) -> "Your shield absorbs most of the dragon's fiery breath!"
+        !success -> "You manage to resist some of the dragon fire!"
+        else -> "You're horribly burnt by the dragon fire!"
     }
 
     internal fun maxHit(type: String, success: Boolean, shield: Boolean, protection: Boolean = false, potion: Int = 0): Int {
--- a/game/src/main/kotlin/content/entity/player/effect/Dragonfire.kt
+++ b/game/src/main/kotlin/content/entity/player/effect/Dragonfire.kt
@@ -44,7 +44,7 @@ object Dragonfire {
         } else if (source is Player) {
             return SHIELD_MAX_HIT
         } else if (target is Player) {
-            target.message(message(target, success), ChatType.Filter)
+            target.message(message(target), ChatType.Filter)
         }
         return maxHit(
             type = type,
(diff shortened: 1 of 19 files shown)


game/src/main/kotlin/content/entity/player/effect/Dragonfire.kt
```diff
@@ -32,12 +39,15 @@ object Dragonfire {
...
             }
         } else if (source is Player) {
-            type = "chromatic"
+            return SHIELD_MAX_HIT
         } else if (target is Player) {
-            target.message(if (success) "You're horribly burnt by the dragon fire!" else "You manage to resist some of the dragon fire!")
+            if (Equipment.antiDragonShield(target)) {
+                target.message("Your shield absorbs most of the dragon's fiery breath!", ChatType.Filter)
+            }
+            target.message(if (success) "You're horribly burnt by the dragon fire!" else "You manage to resist some of the dragon fire!", ChatType.Filter)
```
GregHib: Are these two messages meant to be possible to be sent at the same time?

HarleyGilpin: No — good catch, that was a bug. A shielded player got both "Your shield absorbs most of the dragon's fiery breath!" and "You're horribly burnt by the dragon fire!", which contradict each other.

Both references pick exactly one message by protection level rather than stacking them — 2009scape's `getDragonfireMessage` ladders shield+potion → shield → potion → prayer → burnt, and Darkan returns early per protection source. Reduced to a single `message()` picking the shield line when one is worn, otherwise the existing resist/burnt pair.

The test for this now also asserts the burnt message is *not* sent alongside it.

Fixed in 88627bdd5

HarleyGilpin: Followed this up with full ladder in 3d70af06c

| state | message |
| --- | --- |
| super antifire | Your potion heavily protects you from the dragon's fire. |
| shield + potion | Your shield and potion fully protect you from the heat of the dragon's breath. |
| shield | Your shield manages to block some of the dragon's breath. |
| potion | Your potion slightly protects you from the heat of the dragon's breath. |
| prayer | Your prayers help resist some of the dragonfire! |
| nothing | You are hit by the dragon's fiery breath. |

GregHib (approved):

PR #1188, opened 2026-08-20
Blessed spirit shield creation and Prayer level check fixes

HarleyGilpin:
## Blessed spirit shield creation

Using a holy elixir on a spirit shield at an altar of Saradomin now blesses it, taking level 85 Prayer and granting 1500 Prayer experience. Brother Jered will do the same for a 1,000,000 coin donation to the monastery, with no Prayer requirement and no experience, matching the wiki's two creation routes.

Jered's dialogue was registered against `brother_jared` while the npc is `brother_jered`, so none of it ran; his holy symbol blessing and Prayer skillcape offer were both unreachable. The script is renamed to match the npc.

Summer's End has no script yet, so the quest requirement is left off rather than making the content unreachable, matching the spirit shield drops which are already ungated.

## Prayer level checks

`Level.has` reads the current level, which for Prayer is the player's remaining prayer points, so three requirements were really asking for unspent points:

- Abbot Langley refused to let anyone join the monastery order once they had prayed at all, whatever their Prayer level. The check was duplicated in the conversation and on the ladder, so it now lives in `joinMonasteryOrder`, and joining at the ladder carries on up rather than cancelling the climb and making the player click it a second time.
- Repairing a gravestone (2 Prayer) and blessing one (70 Prayer) were locked the same way. Blessing keeps its separate check that the player has prayer points left to spend, which is the one place the current level is the right thing to read.

All four now test the base level.

## Testing

- `SpiritShieldBlessingTest` - blessing at an altar, the 85 Prayer requirement, and the missing elixir case
- `BrotherJeredTest` - blessing for the donation, and refusing without the coins
- `AbbotLangleyTest` - joining with prayer points spent, refusal below 31, and climbing the ladder after joining
- `GravestonesTest` - existing repair and bless tests now set experience rather than the current level, with the repairer drained to no prayer points, plus a new case pinning the points check on blessing

### DIFF
--- a/game/src/main/kotlin/content/area/misthalin/edgeville/monastery/BrotherJered.kt
+++ b/game/src/main/kotlin/content/area/misthalin/edgeville/monastery/BrotherJered.kt
@@ -8,17 +8,60 @@ import world.gregs.voidps.engine.entity.character.player.skill.Skill
 import world.gregs.voidps.engine.entity.character.player.skill.level.Level.hasMax
 import world.gregs.voidps.engine.inv.inventory
 import world.gregs.voidps.engine.inv.replace
+import world.gregs.voidps.engine.inv.transact.operation.AddItem.add
+import world.gregs.voidps.engine.inv.transact.operation.RemoveItem.remove
 
-class BrotherJared : Script {
+class BrotherJered : Script {
+
+    private val cost = 1_000_000
+
     init {
-        npcOperate("Talk-to", "brother_jared") { (target) ->
+        npcOperate("Talk-to", "brother_jered") { (target) ->
             choice {
                 bold()
+                if (inventory.contains("holy_elixir") && inventory.contains("spirit_shield")) {
+                    option<Quiz>("Can you do anything with this holy elixir?") {
+                        offerBlessing()
+                    }
+                }
                 option<Happy>("Praise be to Saradomin!") {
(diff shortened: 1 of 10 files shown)


game/src/main/kotlin/content/area/misthalin/edgeville/monastery/BrotherJered.kt
```diff
@@ -8,19 +8,62 @@ import world.gregs.voidps.engine.entity.character.player.skill.Skill
...
+        }
+        val success = inventory.transaction {
+            remove("holy_elixir")
+            remove("spirit_shield")
+            remove("coins", cost)
+            add("blessed_spirit_shield")
+        }
+        if (!success) {
+            return
+        }
```
GregHib:
```suggestion
        val success = inventory.transaction {
            remove("holy_elixir")
            remove("spirit_shield")
            remove("coins", cost)
            add("blessed_spirit_shield")
        }
        if (!success) {
            player<Sad>("But, unfortunately, I don't have enough money with me.")
            npc<Neutral>("I am sorry to hear that. If you should find yourself in wealthier times come back and see me.")
            return
        }
```

GregHib (changes requested):

GregHib (approved):

PR #1193, opened 2026-08-20
Spirit shield sigil attaching

HarleyGilpin:
Smithing one of the corporeal beast's four sigils onto a blessed spirit shield at an anvil now makes the spirit shield of that sigil's kind, taking 85 Smithing, base level 90 Prayer and a hammer for 1800 Smithing experience. Either the sigil or the shield can be the item used on the anvil; using the shield picks up whichever sigil is carried.

Prayer is checked against the base level because the current level in this engine is the player's remaining prayer points, so praying at all would otherwise lock a player out of their own shield.

Brother Bordiss will do the same for 1,500,000 coins. He had no dialogue at all before this, so he also answers for himself on what a dwarf is doing in a Saradominist monastery.

Perils of Ice Mountain has no script yet, so Bordiss's quest requirement is left off rather than making the service unreachable, matching how the blessing and the sigil drops are already ungated.

### DIFF
--- a/game/src/main/kotlin/content/area/misthalin/edgeville/monastery/BrotherBordiss.kt
+++ b/game/src/main/kotlin/content/area/misthalin/edgeville/monastery/BrotherBordiss.kt
@@ -59,10 +59,6 @@ class BrotherBordiss : Script {
             npc<Neutral>("You'll need a blessed spirit shield for me to mount it on. Come back and see me once you have one.")
             return
         }
-        if (!inventory.contains("coins", cost)) {
-            player<Sad>("I don't seem to have enough coins, I will return once I do.")
-            return
-        }
         val shield = SpiritShieldSigils.shield(sigil)
         val success = inventory.transaction {
             remove(sigil)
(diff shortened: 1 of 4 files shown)


game/src/main/kotlin/content/area/misthalin/edgeville/monastery/BrotherBordiss.kt
```diff
@@ -0,0 +1,79 @@
...
+        val shield = SpiritShieldSigils.shield(sigil)
+        val success = inventory.transaction {
+            remove(sigil)
+            remove(SpiritShieldSigils.SHIELD)
+            remove("coins", cost)
+            add(shield)
+        }
+        if (!success) {
+            return
+        }
```
GregHib:
```suggestion
        val shield = SpiritShieldSigils.shield(sigil)
        val success = inventory.transaction {
            remove(sigil)
            remove(SpiritShieldSigils.SHIELD)
            remove("coins", cost)
            add(shield)
        }
        if (!success) {
            player<Sad>("I don't seem to have enough coins, I will return once I do.")
            return
        }
```

GregHib (approved):

PR #1194, opened 2026-08-20
Add Evil Chicken's Lair

HarleyGilpin:
Using a raw chicken on the chicken shrine north-east of Zanaris now teleports into the Evil Chicken's Lair, and the portal inside teleports back out to the shrine.

A rope used on the tunnel entrance on the west wall opens it for three minutes, leading down to the baby black dragon. The permanent rope at the bottom climbs back up. The tunnel entrance isn't named like a rope, so the generic climb animation for ladders doesn't cover it and a climb-down animation is registered for the roped variant.

Both teleports use the disappear and reappear pair, declared as a single tick each: the animations run 40 client ticks, so rounding up to two ticks let the player stand back up before the teleport fired.

K'klik is spawned by the nest, and Chickened Out plays in the lair.

[Screencast_20260820_134931.webm](https://github.com/user-attachments/assets/fa3d428a-6f33-4d94-9cd9-4987183136a3)

### DIFF
--- a/game/src/main/kotlin/content/area/misthalin/zanaris/evil_chicken_lair/EvilChickenLair.kt
+++ b/game/src/main/kotlin/content/area/misthalin/zanaris/evil_chicken_lair/EvilChickenLair.kt
@@ -21,7 +21,7 @@ class EvilChickenLair : Script {
                 return@itemOnObjectOperate
             }
             message("You place the raw chicken on the shrine.")
-            delay(anim("teleport_evil_chicken_lair"))
+            animDelay("teleport_evil_chicken_lair")
             tele(ENTRANCE)
             anim("teleport_land_evil_chicken_lair")
         }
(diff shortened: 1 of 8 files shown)


game/src/main/kotlin/content/area/misthalin/zanaris/evil_chicken_lair/EvilChickenLair.kt
```diff
@@ -0,0 +1,65 @@
...
+
+class EvilChickenLair : Script {
+    init {
+        itemOnObjectOperate("raw_chicken", "chicken_shrine") {
+            walkToDelay(SHRINE_TILE)
+            if (!inventory.remove("raw_chicken")) {
+                return@itemOnObjectOperate
+            }
+            message("You place the raw chicken on the shrine.")
+            delay(anim("teleport_evil_chicken_lair"))
```
GregHib:
```suggestion
            animDelay("teleport_evil_chicken_lair")
```

GregHib (approved):
Bwuk bwuk bwuk!

PR #1196, opened 2026-08-21
Bonus Prayer experience at Wilderness chaos altars

HarleyGilpin:
Offering bones at an altar now grants 50% more Prayer experience than burying them. The chaos altars in the Chaos Temple north of Varrock and the Chaos Temple hut in level 38 Wilderness grant 250% extra instead.

### Implementation

The bonus is data driven, in `data/skill/prayer/bone_offering.tables.toml` keyed by object id, defaulting to 50% for altars without a row.

The Chaos Temple hut altar is object 411, which is also used at Ourania, the Asgarnia temple and two other locations that shouldn't grant the bonus, so that row is flagged `wilderness_only` and checked against the existing `wilderness` area.

Also included:

- Adds the missing `bone_offering` graphic (624). `BoneOffering` already spawned it on the altar, but no definition existed so nothing rendered.
- Renames object 411 from `prayer_altar_chaos_ourania` to `prayer_altar_chaos`; the id is used at five separate locations, so the Ourania suffix was misleading.

### Testing

Three tests in `PrayerTest`: bones give 15.7 experience at both Wilderness chaos altars and 6.7 at a regular altar, against 4.5 for burying.

### DIFF
--- a/cache/src/main/kotlin/world/gregs/voidps/cache/definition/Params.kt
+++ b/cache/src/main/kotlin/world/gregs/voidps/cache/definition/Params.kt
@@ -261,9 +261,11 @@ object Params {
     const val SUMMONING_BEAST_OF_BURDEN_ESSENCE = 5264
     const val SUMMONING_SPECIAL_COST = 5265
     const val DISENGAGE = 5266
+    const val BONE_OFFERING_BONUS = 5267
 
     private fun custom(name: String) = when (name) {
         "disengage" -> DISENGAGE
+        "bone_offering_bonus" -> BONE_OFFERING_BONUS
         "summoning_beast_of_burden_essence" -> SUMMONING_BEAST_OF_BURDEN_ESSENCE
         "summoning_special_cost" -> SUMMONING_SPECIAL_COST
         "variables" -> VARIABLES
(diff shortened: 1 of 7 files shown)


GregHib (commented):
Why use area? Just put the object id + bonus in a tables.toml

HarleyGilpin:
Done, moved to `data/skill/prayer/bone_offering.tables.toml` keyed by object id - fe99a87dc.

The Chaos Temple hut altar is object 411, which is also used at Ourania, the Asgarnia temple and two other spots that shouldn't get the bonus, so that row carries a `wilderness_only` flag checked against the existing `wilderness` area.

M1krofin:
Hmm, was this authentic back in 2011? I checked that Jagex added 2016 bonus xp to chaos altar hut with achievement diary but I can be wrong. Not so much data from Jagex part of things 😄

HarleyGilpin:
> Hmm, was this authentic back in 2011? I checked that Jagex added 2016 bonus xp to chaos altar hut with achievement diary but I can be wrong. Not so much data from Jagex part of things 😄

It's hard to tell if it was a rework or not. At the time I believe everyone was using construction (gilded altars) to train prayer, I wasn't aware of any methods of using chaos altar in wilderness as a training method. I don't see any old guides referencing it either. I know there was a 50% bonus at normal altars, however.

HarleyGilpin:
Confirmed via old Fandom members training guide that there was no method for Chaos Altar in 2011: https://runescape.fandom.com/wiki/Members_Prayer_training?oldid=3875617

PR #1200, opened 2026-08-21
Add pickable banana trees on Karamja

HarleyGilpin:
## What

The Musa Point plantation banana trees (objects 2073-2078, 32 of them in region 11569) were unnamed scenery, so their `Pick` option did nothing, and Luthas had no dialogue.

### Picking

- Names the five-stage chain `karamja_banana_tree_5` … `karamja_banana_tree_0` in `musa_point.objs.toml`. `banana_tree_1..5` was already taken by the farming fruit tree patch (7994-7998), hence the prefix.
- Adds `BananaTrees` - picking gives a banana, plays anim 832 and sound 2581, prints "You pick a banana.", and steps the tree down one stage. The tree regrows to full after `world.objs.bananaTree.regrowTicks` (300 by default).
- Object 2078 (empty tree) carries a `Search` option rather than `Pick`, so it gets its own "There are no bananas left on the tree." handler. Note 2078 is also placed as jungle scenery in regions 12844/13100/13101, where that message reads correctly too.
- Hooks the previously unused Karamja easy task **Five a day** (varbit 3566) with a new persisted `five_a_day_progress` counter.

### Luthas and the export crate

- Adds `Luthas` dialogue. The opening choice is either asking for employment on the plantation, or the conversation about the customs officer who no longer searches his export crates and the shop in Port Sarim they get shipped to.
- Taking the job sets `banana_plantation_job`. While employed, Luthas asks whether the task is done; once the crate holds ten bananas he pays 30 coins and clears both the counter and the job. Payment aborts cleanly on a full inventory rather than emptying the crate for nothing.
- The crate supports `Search`, `Fill`, and banana-on-crate, capacity 10, anim 832 on each. Without the job every interaction answers "I don't know what goes in there."
- Object 2072 is the shared `crate_14` definition and is also placed on Ape Atoll (2746, 2768), so the handlers are gated to the plantation crate's tile to leave that one inert.

Rum smuggling is deliberately left out - Pirate's Treasure isn't implemented, so stashing rum in the crate would be a dead end.

## Why the regrow isn't a plain replace chain

`GameObjects.replace(original, replacement, ticks)` registers one timer over `setOf(original, replacement)` whose revert re-adds the *immediately previous* object, and `GameObjectTimers.add` cancels any timer touching either object. Chained across five stages, a rapidly-picked tree reverts one stage and is then left with no timer at all - permanently stranded mid-chain.

Since `GameObjectHashMap.add(obj, REPLACED)` only ORs a flag and never overwrites the stored map value, the original is always recoverable. So each pick replaces permanently and re-arms a single timer that removes the replacement, reverting straight back to the full tree.

## Testing

`BananaTreesTest` covers picking one banana, emptying a tree (and the diary task firing), the empty-tree message, full regrowth after the timer, and a full inventory being rejected.

`LuthasTest` covers both dialogue branches, taking the job, filling the crate, packing a single banana, getting paid, and the crate staying inert without the job.

`BushPickingTest`, `ObjectTest` and `CustomsOfficerTest` still pass.

### DIFF
--- a/game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
+++ b/game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
@@ -19,7 +19,7 @@ class BananaTrees : Script {
                 return@objectOperate
             }
             sound("pick")
-            anim("climb_down")
+            anim("take")
             message("You pick a banana.")
 
             val remaining = target.id.removePrefix("karamja_banana_tree_").toInt() - 1
--- a/game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
+++ b/game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
@@ -5,7 +5,6 @@ import world.gregs.voidps.engine.client.message
 import world.gregs.voidps.engine.data.Settings
 import world.gregs.voidps.engine.entity.character.player.chat.inventoryFull
 import world.gregs.voidps.engine.entity.character.sound
-import world.gregs.voidps.engine.entity.obj.GameObjects
 import world.gregs.voidps.engine.entity.obj.replace
 import world.gregs.voidps.engine.inv.add
 import world.gregs.voidps.engine.inv.inventory
@@ -23,16 +22,11 @@ class BananaTrees : Script {
             message("You pick a banana.")
 
             val remaining = target.id.removePrefix("karamja_banana_tree_").toInt() - 1
-            // Replaced permanently with a single re-armed timer; removing the replacement reverts
-            // the tree all the way back to the original full one rather than only one stage.
-            GameObjects.timers.cancel(target)
-            val replacement = target.replace("karamja_banana_tree_$remaining")
-            GameObjects.timers.add(replacement, Settings["world.objs.bananaTree.regrowTicks", 300]) {
-                GameObjects.remove(replacement, collision = true)
-            }
+            target.replace("karamja_banana_tree_$remaining", ticks = Settings["world.objs.bananaTree.regrowTicks", 300])
 
             if (!get("five_a_day_task", false) && inc("five_a_day_progress") >= 5) {
                 set("five_a_day_task", true)
+                clear("five_a_day_progress")
             }
         }
 
--- a/game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
+++ b/game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
@@ -1,0 +1,43 @@
+package content.area.karamja.musa_point
+
+import world.gregs.voidps.engine.Script
+import world.gregs.voidps.engine.client.message
+import world.gregs.voidps.engine.data.Settings
+import world.gregs.voidps.engine.entity.character.player.chat.inventoryFull
+import world.gregs.voidps.engine.entity.character.sound
+import world.gregs.voidps.engine.entity.obj.GameObjects
+import world.gregs.voidps.engine.entity.obj.replace
+import world.gregs.voidps.engine.inv.add
+import world.gregs.voidps.engine.inv.inventory
+
+class BananaTrees : Script {
+
+    init {
+        objectOperate("Pick", "karamja_banana_tree_1,karamja_banana_tree_2,karamja_banana_tree_3,karamja_banana_tree_4,karamja_banana_tree_5") { (target) ->
+            if (!inventory.add("banana")) {
+                inventoryFull()
+                return@objectOperate
+            }
+            sound("pick")
+            anim("climb_down")
+            message("You pick a banana.")
+
+            val remaining = target.id.removePrefix("karamja_banana_tree_").toInt() - 1
+            // Replaced permanently with a single re-armed timer; removing the replacement reverts
+            // the tree all the way back to the original full one rather than only one stage.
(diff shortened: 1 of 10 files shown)


game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
```diff
@@ -0,0 +1,43 @@
...
+            // Replaced permanently with a single re-armed timer; removing the replacement reverts
+            // the tree all the way back to the original full one rather than only one stage.
+            GameObjects.timers.cancel(target)
+            val replacement = target.replace("karamja_banana_tree_$remaining")
+            GameObjects.timers.add(replacement, Settings["world.objs.bananaTree.regrowTicks", 300]) {
+                GameObjects.remove(replacement, collision = true)
+            }
+
+            if (!get("five_a_day_task", false) && inc("five_a_day_progress") >= 5) {
+                set("five_a_day_task", true)
```
GregHib, line 28: Can probably clear the `five_a_day_progress` after completion

game/src/main/kotlin/content/area/karamja/musa_point/Luthas.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+            val carrying = inventory.count("banana")
+            if (carrying == 0) {
+                message("You don't have any bananas to pack.")
+                return@objectOperate
+            }
+            val packed = minOf(space, carrying)
+            if (!inventory.remove("banana", packed)) {
+                return@objectOperate
+            }
```
GregHib:
```suggestion
            val packed = inventory.removeToLimit("banana", space)
            if (packed == 0) {
                message("You don't have any bananas to pack.")
                return@objectOperate
            }
```

game/src/main/kotlin/content/area/karamja/musa_point/Luthas.kt
```diff
@@ -0,0 +1,148 @@
...
+            animDelay("take")
+            inc("banana_crate_bananas", packed)
+            if (packed == carrying) {
```
GregHib:
```suggestion
            if (!inventory.contains("banana")) {
```

game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
```diff
@@ -0,0 +1,43 @@
...
+            message("You pick a banana.")
+
+            val remaining = target.id.removePrefix("karamja_banana_tree_").toInt() - 1
+            // Replaced permanently with a single re-armed timer; removing the replacement reverts
+            // the tree all the way back to the original full one rather than only one stage.
+            GameObjects.timers.cancel(target)
+            val replacement = target.replace("karamja_banana_tree_$remaining")
+            GameObjects.timers.add(replacement, Settings["world.objs.bananaTree.regrowTicks", 300]) {
+                GameObjects.remove(replacement, collision = true)
+            }
```
GregHib: I think just doing this should be fine? Objects revert to their original not their replacements which is what happens on osrs at least, it doesn't grow back 1 banana at a time.
```suggestion
            target.replace("karamja_banana_tree_$remaining", ticks =  Settings["world.objs.bananaTree.regrowTicks", 300])
```

GregHib (changes requested):

game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
HarleyGilpin: You're right that this is what it should do, but the engine wasn't doing it for a chain of replacements, so I've fixed that instead of working around it in content.

`replace`'s revert block did `remove(replacement); add(original)`, where `original` is whatever object was passed in, for the second pick that's the *previous replacement*, not the map original. `GameObjectTimers.add` also cancels any timer touching either object, so the earlier revert is dropped. Picking three bananas and waiting the 300 ticks left the tree at `karamja_banana_tree_3` with no timer at all, stranded there permanently.

`GameObjectHashMap.add(obj, REPLACED)` only ORs the flag and never overwrites the stored value, so `remove(replacement)` already re-adds the original by itself. The fix is to skip the `add(original)` when the replacement is on the same tile and there is an original to fall back on, different-tile replacements (doors) and objects with no original still need it:

```kotlin
val restored = replacement.index == original.index && map[original] > 1
```

With that in place your suggestion works, so the picking code is now just the one line. It also fixes the same latent issue in `BushPicking` (cadava full → half → empty previously reverted only to half). Added `Replacing a replacement is undone back to the original` to `GameObjectsTest`; full suite passes.

game/src/main/kotlin/content/area/karamja/musa_point/BananaTrees.kt
HarleyGilpin: Done, cleared alongside setting the task.

game/src/main/kotlin/content/area/karamja/musa_point/Luthas.kt
HarleyGilpin: Applied.

game/src/main/kotlin/content/area/karamja/musa_point/Luthas.kt
HarleyGilpin: Applied. Pairs with removeToLimit above, since the carried count is no longer needed.

GregHib (approved):

PR #1202, opened 2026-08-23
Rat activity for bots

PrettyPrincessPotato:
Added a kill_rats template to the combat.template.toml
Added kill_lumbridge_castle_rats_* for attack, defence, and strength.

Made the limit 2, though from early testing it might be able to be raised to 3 with little issues.


GregHib (approved):
Very nice thanks!

PR #1205, opened 2026-08-23
H.A.M. Hideout feature-complete

V0lcanic:
A feature-complete H.A.M. Hideout with a few other adjustments and fixes. The H.A.M. storerooms were not added due to the quest "Death to the Dorgeshuun" not being implemented yet. There have also been changes to the engine after discovering bugs that affected the hideout and general gameplay; these are explained in more detail below.

Important changes worth mentioning:

-Fixed equipped item desync by making appearance updates re-send whenever an observer’s cached appearance hash no longer matches the player’s real current appearance, which makes equip/unequip changes stay correct even after ladder/view changes.
-Forced newly added local players to send their face direction on initial visibility, which stopped the client from defaulting them to east.
-Added support for sending private zone updates to only the owning player.
-Male and female H.A.M. members now behave like distinct 2011-style pickpocket targets in the generic thieving system while still sharing the intended HAM hideout-specific fail/punishment mechanics.
-Pickpocketing now uses adjacent NPC operate interaction rather than a looser approach interaction.
• The previous approach-based interaction allowed pickpocket attempts from too far away in obstacle-heavy spaces.
• For HAM specifically, that let players interact across the prison barrier if they got close enough.
• Converting to normal adjacent interaction makes it behave better.

Sources: osrs.world, osrs.world 634, osrs live server, rs3 live server, runescape.wiki, oldschool.runescape.wiki, https://www.youtube.com/watch?v=FMFw3U96l2A, https://www.youtube.com/watch?v=DxGwXDtBK64, https://www.youtube.com/watch?v=YNhqG0uUe14, https://www.youtube.com/watch?v=HZDKTbUpKZs, https://www.youtube.com/watch?v=FYRQ_t6pusM, https://www.youtube.com/watch?v=CK8xh0ifsvw, https://www.youtube.com/watch?v=BC0FR-9BxSI, https://www.youtube.com/watch?v=ImeMs9P5M3w, https://www.youtube.com/watch?v=ECbmNALa4OE

Some fixes:

**Pickpocketing.kt**
[What was going wrong?]
Pickpocketing was being treated like a loose approach interaction instead of a true adjacent NPC interaction.
• Pickpocket was registered with npcApproach("Pickpocket") / npcApproach("Steal-from").
• Inside that flow, it explicitly used approachRange(2).
• That meant the engine would allow the action once the player was within 2 tiles and had a valid approach state, instead of requiring the player to be properly adjacent and actually able to interact with the NPC.
• In cramped HAM hideout spaces, especially around the prison door/gates, this let players trigger pickpocket attempts when they were not truly in legal reach.

[What changed?]
• Changed the interaction from:
◦ approach-based pickpocketing
• to:
◦ operate-based pickpocketing
So it now uses:
• npcOperate("Pickpocket")
• npcOperate("Steal-from")
instead of the old npcApproach(...) calls.

[Why that fixed it]
• npcOperate uses the engine’s normal close-proximity interaction rules.
• That means the player has to actually get into proper adjacent reach of the NPC.
• If a closed door, gate, or impossible path blocks that interaction, the pickpocket does not fire.
• So the fix was essentially: stop treating pickpocket as a ranged/loose “approach” action, and make it a strict adjacent operation.

**PlayerUpdateTask.kt**
[What was going wrong?]
• Player 2 went down the ladder first and was already standing underground, facing south-west.
• When player 1 came down after, player 2 was being added as a new local player on player 1’s client.
• On that “new local player” add, the client got position/appearance, but it was not reliably getting the current face direction.
• So the client fell back to its default render direction, which showed up as east on player 1’s screen.

[What changed?]
• In PlayerUpdateTask.kt, the initial add flag for newly visible players includes:
◦ PLAYER_FACE_MASK
◦ along with appearance / movement-type / temp-move-type / animation
That means when a player is first introduced into another player’s local view, the server now sends their current facing direction immediately, instead of waiting for some later turn/update packet.

[Why that fixed it]
• The bug was a spawn-in-view state sync issue, not a pathing issue.
• Player 2’s real direction underground was already correct server-side.
• The missing piece was making sure that direction was included in the first update packet seen by player 1 after entering.

**PlayerUpdateTask.kt**
[What was going wrong?]
Going down the ladder changes who is in whose local view and can temporarily break the normal “live appearance update” flow.
If player 1 changed equipment after that, the server-side appearance changed correctly, but player 2 could still keep the old cached appearance.
That meant player 2 would still see the old worn items until some later full refresh happened.

[What changed?]
1. Appearance changes are hashed when flagged
In PlayerVisuals.kt, flagAppearance() recalculates:
◦ appearance.hash
◦ appearance.length
2. Each observer tracks the last appearance hash they were sent
In PlayerTrackingSet.kt, each tracked player has a cached appearanceHash.
3. The update task now forces an appearance update if the hashes differ
In PlayerUpdateTask.kt, updateFlag() checks:
◦ observer’s cached appearance hash
◦ target player’s current appearance hash
If they do not match, it forces APPEARANCE_MASK into the outgoing update even if there wasn’t some other obvious movement/update reason to resend appearance.
4. After the appearance packet is sent, the observer cache is updated
set.updateAppearance(player) stores the new hash so future updates stay in sync.

[Why the ladder exposed it]
• The HAM ladder/hideout flow changes visibility/local-player state, so it was a good place for the stale cached appearance bug to show up.
• The real issue was: appearance sync depended too much on the exact tick the equipment change happened.
The hash comparison made it state-based instead of timing-based.

Co-authored-by: Copilot <copilot@local>


engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/npc/NPC.kt
```diff
@@ -32,6 +32,9 @@ data class NPC(
     override val visuals: NPCVisuals = NPCVisuals()
 
     var hide = false
+    private val collidesWithNpcs: Boolean
+        get() = this["collides_with_npcs", true]
+
```
GregHib, line 37: For what purpose is this added where `solid = false` isn't enough?

game/src/main/kotlin/content/area/misthalin/ham_hideout/HamDeacon.kt
```diff
@@ -0,0 +1,183 @@
...
+    companion object {
+        private const val PREACH_QUEUE = "ham_deacon_preach_sequence"
+        private const val PREACH_DELAY_MIN = 3
+        private const val PREACH_DELAY_MAX = 5
+        private const val HAM_HIDEOUT_REGION_ID = 12694
+        private const val POST_SERMON_DELAY = 2
+        private const val POST_SERMON_STAGE_DELAY = 2
+        private val PREACH_EMOTES = listOf("emote_angry", "emote_cheer", "emote_preach")
+
+        // Paraphrased to avoid using verbatim game dialogue.
```
GregHib, line 26: We want to use verbatim game dialogue 😅

GregHib (changes requested):
RuneScape in 2011 didn't have the concept of private object changes, either the trapdoor change occurs for all players, or like enterDoor() it's a quick visual change which is reverted a tick or two later. Your AI has engineered a whole new system around this misunderstanding, so I'll have to decline this PR on that basis.

There's also a number of other broad issues, like over-use of constants for data which should be stored in tables.toml files, and static methods like in HamHideout.kt which seem over-engineered and generally done in a more java-y way.

It'd take a lot of back-and-forth to get this large PR ship-shape so instead if you could split it up into smaller PR parts so I can review them individually that would be great

V0lcanic:
> RuneScape in 2011 didn't have the concept of private object changes, either the trapdoor change occurs for all players, or like enterDoor() it's a quick visual change which is reverted a tick or two later. Your AI has engineered a whole new system around this misunderstanding, so I'll have to decline this PR on that basis.
> 
> There's also a number of other broad issues, like over-use of constants for data which should be stored in tables.toml files, and static methods like in HamHideout.kt which seem over-engineered and generally done in a more java-y way.
> 
> It'd take a lot of back-and-forth to get this large PR ship-shape so instead if you could split it up into smaller PR parts so I can review them individually that would be great

Me and Copilot tried to cook and burned the entire hideout as a result.

In seriousness, I've decided to pivot back to focusing on the general 2011 npc/item spawns, examine texts, dialogues, and making slight adjustments across the game. That is something I have confidence in, instead of trying to add or change entire systems using pure AI code.

I'm not familiar with writing or reading code, so trying to explain most of the lines isn't possible. In a way, this unintentionally devalues the project while taking away time and resources, which isn't fair to you or other contributors.

GregHib:
> Me and Copilot tried to cook and burned the entire hideout as a result.

> In seriousness, I've decided to pivot back to focusing on the general 2011 npc/item spawns, examine texts, dialogues, and making slight adjustments across the game. That is something I have confidence in, instead of trying to add or change entire systems using pure AI code.

> I'm not familiar with writing or reading code, so trying to explain most of the lines isn't possible. In a way, this unintentionally devalues the project while taking away time and resources, which isn't fair to you or other contributors.

That's alright, there's no harm diving a little into code in small amounts, some of these dialogues and interactions (like ButtonPolishing) were fine. If you find bigger problems feel free to just create bug issues and leave it to someone else to address

PR #1213, opened 2026-08-25
Fix compost bins disappearing after relogin

HarleyGilpin:
## Bug

Fill a compost bin, close it, log out and back in, the bin object vanishes from the world permanently.

## Root cause

Compost bins are client-side loc transforms driven by the persisted string varbits 740–743. On login the farming tick replays every offline minute in one burst, fast-forwarding a closed bin through all rot stages. Rot completion in `Farming.growCompost` did `value.replace(stage, "ready")`, producing `compostable_rotting_ready` — a key that doesn't exist in the varbit value map. `MapValues.toInt` silently returns -1 for unknown keys, so the client receives an invalid transform index and stops rendering the bin. Because the variable is `persist = true`, the bad value is saved and re-sent on every later login, so the bin stays gone.

## Fixes

- `Farming.growCompost` / `::rot` command: build the ready value as `substringBefore("_rotting") + "_ready"` so it lands on a real map key.
- `compost_bin.varbits.toml`: add the missing `supercompostable_rotting_1` to all four bins (closing a full supercompost bin had a 1-in-15 chance of instant vanish). Verified against the cache: loc 7836 has no transform at index 63 (64–126 all render the closed bin 7813), so it aliases to 64 like `rotting_0`.
- `CompostBin` Close pattern narrowed to `compostable`/`supercompostable`/`tomatoes` — the old `compost_bin_*_15` wildcard also matched finished `compost_15`/`supercompost_15`/`rotten_tomatoes_15` states and produced invalid `compost_rotting_N` values.
- `CompostBin` playerSpawn: repairs saves already corrupted with `*_rotting_ready` so affected players recover on next login.
- Engine: `VariableDefinition.send` now logs a warning and skips sending when a map variable's value has no entry, instead of shipping -1 to the client — future bad values show up as a log line rather than a silently vanishing object.

## Tests

- `CompostBinTest`: offline catch-up completes to `*_ready`, finished bins can't be closed, every reachable rotting stage (0–30) has a varbit value, corrupted saves repaired on spawn.
- `VariableDefinitionTest`: unknown map value is not sent.

Note: `FarmingTest > Crop progresses through growth stages` currently fails on main as well — pre-existing, unrelated to this change.


data/skill/farming/compost_bin.varbits.toml
GregHib, line 1: Is this all just formatting?

GregHib (approved):

data/skill/farming/compost_bin.varbits.toml
HarleyGilpin: supercompostable_rotting_1 = 64 was missing from all 4 bins.

HarleyGilpin:
Pushed 33180b16 fixing the failing `FarmingTest > Compost ready after 30 rot stages`, it was asserting the old corrupted `compostable_rotting_ready` value. The workflow run on the new commit needs approval when you get a chance.

PR #1216, opened 2026-08-26
Stop run energy restoring while running

### LINKED ISSUE #1214: Run energy recovers while running
**Description**
Run energy seems to be recovering while the player is running. At sufficiently high agility and low carry weight, the result will be infinite run energy!

**To Reproduce**
1. Set Agility to 1, carry 0kg
2. Run: observe how run energy depletes
3. Set Agility to 40, carry 0kg
4. Run: observe how run energy does not deplete past 99%
5. Keep Agility at 40, carry >25kg (e.g. fill inventory with iron ore)
6. Run: run energy should now at least somewhat deplete while running

**Expected behavior**
Run energy recovery should either be slower or (more likely) run energy should not recover at all when the player is actively running.

**Desktop:**
- OS: Nobara Linux 44 (KDE Plasma Desktop Edition) x86_64 with kernel Linux 7.2.0-202.nobara.fc44.x86_64

**Additional context**
https://runescape.wiki/w/Energy?oldid=3597013

**Unrelated, but still important**: there's currently no real ability for people to make "confidental" issues. The e-mail address [greghib@users.noreply.github.com](mailto:greghib@users.noreply.github.com) does not seem to be a valid address which can receive e-mail (as seen in https://github.com/GregHib/void/blob/main/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report)
```
<[greghib@users.noreply.github.com](mailto:greghib@users.noreply.github.com)>: Host or domain name not found. Name service
    error for name=users.noreply.github.com type=A: Host not found
```

HarleyGilpin:
Fixes #1214

## Problem

The `energy_restore` soft timer fires every game tick and restored energy regardless of movement state. While running, both restore and drain applied on the same tick, so the observed drain rate was `restore - drain`. Restore scales with Agility (27 to 157 per tick, in hundredths of a percent) while drain is flat 67 per tick at 0kg, so from roughly Agility 50 upwards running was free and energy never depleted.

## Changes

- `Energy.kt`: skip the passive restore on the tick after a run step drained energy. `last_energy_drain` is already stamped once per tick by the drain hook, and only when an actual run step was taken, so a player standing still with run toggled on still recovers as expected. Per the [2011 wiki](https://runescape.wiki/w/Energy?oldid=3597013), energy only recovers while walking or standing, never while running.
- `Energy.kt`: read the `weight` variable as `Double` to match what `Weight.kt` stores. It was previously read with an `Int` default through the erased-generic variable getter, so carried weight never increased drain correctly.

## Tests

Added to `PlayerTest`:
- Energy doesn't restore while running (99 Agility, exact drain of 67 per tick)
- Energy restores again after running stops
- Weight increases energy drain (28 iron ore, ~63kg, exact drain of 132 per tick)

### DIFF
--- a/game/src/main/kotlin/content/entity/player/effect/energy/Energy.kt
+++ b/game/src/main/kotlin/content/entity/player/effect/energy/Energy.kt
@@ -47,6 +47,10 @@ class Energy : Script {
             if (runEnergy >= MAX_RUN_ENERGY) {
                 return@timerTick Timer.CANCEL
             }
+            // Energy only recovers when not running; drains the tick before restores run
+            if (get("last_energy_drain", -1) == GameLoop.tick - 1) {
+                return@timerTick Timer.CONTINUE
+            }
             runEnergy += getRestoreAmount(this)
             return@timerTick Timer.CONTINUE
         }
@@ -63,7 +67,7 @@ class Energy : Script {
     }
 
     fun getDrainAmount(player: Player): Int {
-        val weight = player["weight", 0].coerceIn(0, 64)
+        val weight = player["weight", 0.0].toInt().coerceIn(0, 64)
         var decrement = 67 + ((67 * weight) / 64)
         if (player.hasClock("hamstring")) {
             decrement *= 4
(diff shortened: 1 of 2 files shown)


GregHib:
Can we add an infinite run setting while we're at it :D

GregHib (approved):

PR #1219, opened 2026-08-27
Add The Hand in the Sand quest

Ebp90:
Ones issue: Betty should stop interacting with the player and walk to the nearest tile to the table to place the glass on it. Marked as TODO


Ebp90:
Also Betty needs to sell pink dye after the quest.

GregHib (approved):

PR #1220, opened 2026-08-27
Fix drop table crash when item drop rate is above 1.0

HarleyGilpin:
## Problem

With `world.itemDropRate` set above 1.0, killing an NPC whose drops include an `all` type table throws:

```
java.lang.IllegalArgumentException: Cannot coerce value to an empty range: maximum 1 is less than minimum 3.
    at kotlin.ranges.RangesKt___RangesKt.coerceIn(_Ranges.kt:1449)
    at world.gregs.voidps.engine.entity.item.drop.DropTable.random(DropTable.kt:42)
    at world.gregs.voidps.engine.entity.item.drop.DropTable.roll(DropTable.kt:30)
    at content.entity.death.NPCDeath.dropLoot-ivHGwhI(NPCDeath.kt:104)
```

The exception escapes up through `NPCTask` into `GameLoop.tick`, so the kill drops no loot and the rest of that tick's NPC processing is aborted.

## Cause

`DropTable.random` coerces the scaled roll into `[sum(chances), roll]`. For `type = "all"` tables the loader defaults `roll = 1` and each drop's chance defaults to 1, so any all-type table with 2+ drops (e.g. the bones tables in `misc.drops.toml` reached via nested `collect`) produces an empty range, `coerceIn(3, 1)` for a 3-drop table, matching the trace. Any roll-based table misconfigured with `roll < sum(chances)` hits the same throw.

## Fix

- Skip multiplier scaling for `TableType.All`, `collect` returns every drop regardless of the roll, so there is nothing to boost.
- Clamp `minRoll` to the base roll so tables with chances exceeding their roll no longer throw; well-formed roll-based tables behave exactly as before.

## Tests

Two regression tests added to `DropTableTest`: all-type table with multiplier > 1 no longer throws, and a roll-based table with chances exceeding its roll clamps instead of throwing. `:engine:test` drop suites pass (18 + 19 tests, 0 failures).


HarleyGilpin:
Added a load-time warning when a roll-based table's summed chances exceed its roll. Turns out the crash wasn't limited to all-type tables: 79 tables in data currently overflow their roll (e.g. `dark_wizard_low_secondary` 130/128, `ice_troll_male_secondary` 8827/5012, `kree_arra_primary` which is missing `type = "all"` so feathers never drop). In each of these the drops past the roll can never be selected, and before the fix any of them would throw the coerceIn exception whenever `world.itemDropRate` was above 1.0.

HarleyGilpin:
Data cleanup pass done in 5f29da801, all 79 flagged tables fixed, loader now reports zero warnings across 1,335 tables.

Each table was checked against Jan–Mar 2011 runescape.wiki revisions (OSRS wiki for exact fractions where the monster's table survives unchanged) rather than mechanically bumping rolls. The overflows turned out to have real causes:

- **Guaranteed drops modeled as pick-one**, Kree'arra/aviansie bones+feathers, Warriors' Guild animated armour sets, paladin pickpocket (always 80 coins + 2 chaos runes, court summons 1/1000) → `type = "all"`
- **Inverted numerators**, denominator written as the chance: dagannoth medium clue `128` on a /128 table, stronghold goblin herbs `256`
- **Two tables concatenated**, jelly had the OSRS-only Warped Jelly block appended, ice troll male had the mountain-troll table pasted on, chaos elemental's major+minor rolls and Kalphite Queen's guaranteed food roll were merged into one, rock golem contained a literal duplicate block (the deduped union sums to exactly 128)
- **Combined drops double-counted**, bloodveld/jogre "bones + big bones together" rows, lizard's fire-rune/waterskin companion totals, tortoise arrow/cabbage rates
- **Common tier couldn't fit**, strykewyrms/waterfiend/tormented demon had 12–29 items at 32/512; scaled the common tier down to fit while keeping every uncommon/rare/very-rare numerator exact
- **Padding absorbed the rest**, explicit `nothing` entries and GWD coin lines trimmed so /762 unique+shard rates stay canonical

Along the way this also surfaced **duplicate table names silently clobbering each other** (loader is last-write-wins): a stale `monk.drops.toml` copy of the human tables, orphaned bloodveld/gorak files shadowing the maintained ones, doomion wired to the disciple_of_iban table, Gu'Tanoth's rock cake stall named `food_stall_drop_table` (which hid Ape Atoll's banana stall), and a duplicated mugger pair in one file. All resolved; zero duplicate names remain. Might be worth a load-time warning for duplicate table names as a follow-up.

Full engine + game test suites pass.

GregHib (approved):

PR #1221, opened 2026-08-28
Add ancient effigies

HarleyGilpin:
Adds the [Ancient effigies](https://runescape.wiki/w/Ancient_effigies?oldid=3918860) D&D. Item definitions and the 26 tertiary drop-table entries were already in the data, so this wires up the content on top of them.

## Mechanics

- **Investigate**: each effigy offers one of eight fixed skill pairs (Agility/Crafting, Construction/Thieving, Cooking/Firemaking, Fishing/Farming, Fletching/Woodcutting, Herblore/Hunter, Mining/Smithing, Summoning/Runecrafting). Feeding it a skill at level 91/93/95/97 (boosts count, per wiki) grants 15k/20k/25k/30k xp and advances starved → nourished → sated → gorged → dragonkin lamp.
- The offered pair is rolled on first investigate and stored in a persistent per-stage variable, cleared when fed, so players can't re-click to fish for a preferred pair. Items here only carry id + amount, so per-item storage wasn't an option; the one visible quirk is that two effigies of the same stage share a pair.
- **Dragonkin lamp**: any skill via the standard lamp interface, base (unboosted) level, `(lvl³ − 2lvl² + 100lvl) / 20` xp at level 30+, the previous level's xp delta below that. Matches the wiki's worked example (10,740 at level 60). Shows the [transcript](https://runescape.wiki/w/Transcript:Dragonkin_lamp) line "...but at what cost?".
- **Historian Minas**: hands over any effigy for the 5,000 xp antique lamp (skill must be level 50+), using the documented "I'll have my top men look into this right away." / "Who?" / "Top...men..." exchange.
- All effigy and lamp dialogue uses item boxes so the relic/lamp sprite shows in chat.

## Known gaps

- Minas' opening offer lines are placeholders; only the "top men" exchange is documented.
- No assist system support (not implemented in Void) and no daily cap.

## Testing

14 WorldTest cases: every stage boundary, boost success, level-too-low failure, pair persistence across investigations, lamp formula above/below 30 and the Constitution tenths quirk, the Minas exchange, and both antique lamp outcomes. Full `:game:test` sweep passes.

### DIFF
--- a/game/src/test/kotlin/content/activity/ancient_effigies/AncientEffigiesTest.kt
+++ b/game/src/test/kotlin/content/activity/ancient_effigies/AncientEffigiesTest.kt
@@ -11,6 +11,7 @@ import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Test
+import playerOption
 import skipDialogues
 import world.gregs.voidps.engine.entity.character.player.Player
 import world.gregs.voidps.engine.entity.character.player.skill.Skill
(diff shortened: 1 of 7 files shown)


game/src/test/kotlin/content/activity/ancient_effigies/AncientEffigiesTest.kt
```diff
@@ -0,0 +1,247 @@
...
+import skipDialogues
+import world.gregs.voidps.engine.entity.character.player.Player
+import world.gregs.voidps.engine.entity.character.player.skill.Skill
+import world.gregs.voidps.engine.entity.character.player.skill.level.Level
+import world.gregs.voidps.engine.inv.add
+import world.gregs.voidps.engine.inv.inventory
+import world.gregs.voidps.type.Tile
+
+class AncientEffigiesTest : WorldTest() {
+
```
GregHib, line 24: Can we get a test to check if assistance works with effigies (and if not add support)?

GregHib (changes requested):

game/src/test/kotlin/content/activity/ancient_effigies/AncientEffigiesTest.kt
HarleyGilpin: Added in 1036062ea, the level check now accepts the assistant's (boosted) level for any skill they have toggled on, and the investigation xp flows through the existing blocked-experience redirect so it respects the 30k daily cap (excess is lost but the effigy still advances, matching the wiki). Each effigy skill pair contains exactly one of the nine assistable skills, so no changes to the assist skill list were needed. Covered by three new tests: assisted investigation, under-levelled assistant refused, and the cap-overflow case.

GregHib (approved):

PR #1223, opened 2026-08-28
Fix invalid map variable values from farming and unlock command

### LINKED ISSUE #1218: Unknown variable value
Need to investigate, might just be old values on my dev account
```
Unknown value 'incomplete' for map variable 281; not sent.
WARN [VariableDefinition] Unknown value 'true' for map variable 6494; not sent.
WARN [VariableDefinition] Unknown value 'herb_dead_2' for map variable 707; not sent.
WARN [VariableDefinition] Unknown value 'completed' for map variable 293; not sent.
WARN [VariableDefinition] Unknown value 'false' for map variable 5733; not sent.
WARN [VariableDefinition] Unknown value 'compostable_rotting_ready' for map variable 740; not sent.
WARN [VariableDefinition] Unknown value 'true' for map variable 6494; not sent.
WARN [VariableDefinition] Unknown value 'true' for map variable 6494; not sent.
WARN [VariableDefinition] Unknown value 'herb_dead_2' for map variable 707; not sent.
```

HarleyGilpin:
Investigation of #1218. Of the six warnings, two come from write paths still in the code; the rest are stale saved values.

## Live bugs fixed

**Varbit 707, `herb_dead_2` (Catherby fruit tree)**
The diseased-to-dead transition in `Farming.kt` checked `variable.contains("herb")`, which also matches "cat**herb**y". A diseased apple tree at stage 2 in the Catherby patch was written `herb_dead_2`, a value only herb patch varbit maps define, so the varbit was never sent and the patch stopped rendering its real state. The same applied to the Catherby veg and flower patches. Now matches on `"herb_patch"`, which all five herb patch variables contain. Added a regression test: diseased Catherby apple tree now dies to `apple_dead_2`.

**Varbit 6494, `true` (`introducing_explorer_jack_task`)**
The `unlock tasks` command wrote `true` to every `*_task` struct id, but this task is the one `*_task` varbit with `format = "map"` (`unstarted`/`completed`), so the boolean never sent and the warning repeated on every login via the task list refresh. The command now checks the variable definition and writes `"completed"` for map-format tasks. `TaskSystem` already accepts both forms.

**Trollheim herb patch disease immunity**
`Farming.disease()` compared against `"patch_my_arm_herb"` but the variable is named `farming_herb_patch_my_arm`, so the check never matched and the canonically disease-free patch could disease and die. Low impact until My Arm's Big Adventure is implemented, but the fix is one string.

## Stale save data, no code change

- **Varp 281 `incomplete`**: #1007 renamed the `unstable_foundations` value `incomplete` to `started` with no migration.
- **Varp 293 `completed`** and **varbit 5733 `false`**: `big_chompy_bird_hunting` and `myths_of_the_white_lands` got map definitions in #1007; nothing in the codebase ever wrote those values, so they predate the defs on the dev account.
- **Varbit 740 `compostable_rotting_ready`**: already fixed and save-repaired on spawn by #1213; the logged lines predate that deploy.

Farming test suite passes (FruitTreePatchTest 34, HerbPatchTest 47, 0 failures).

Closes #1218

### DIFF
--- a/game/src/main/kotlin/content/entity/player/command/PlayerCommands.kt
+++ b/game/src/main/kotlin/content/entity/player/command/PlayerCommands.kt
@@ -17,6 +17,7 @@ import world.gregs.voidps.engine.Script
 import world.gregs.voidps.engine.client.command.*
 import world.gregs.voidps.engine.client.message
 import world.gregs.voidps.engine.client.ui.open
+import world.gregs.voidps.engine.client.variable.MapValues
 import world.gregs.voidps.engine.client.variable.PlayerVariables
 import world.gregs.voidps.engine.client.variable.hasClock
 import world.gregs.voidps.engine.client.variable.start
@@ -304,7 +305,8 @@ class PlayerCommands(
         if (type == "all" || type == "tasks" || type == "achievements") {
             for (struct in StructDefinitions.definitions) {
                 if (struct.stringId.endsWith("_task")) {
-                    target[struct.stringId] = true
+                    val definition = VariableDefinitions.get(struct.stringId)
+                    target[struct.stringId] = if (definition?.values is MapValues) "completed" else true
                 }
             }
             target.message("All tasks completed.")
--- a/game/src/main/kotlin/content/skill/farming/Farming.kt
+++ b/game/src/main/kotlin/content/skill/farming/Farming.kt
@@ -28,6 +28,13 @@ class Farming(
 
     init {
         playerSpawn {
+            // Repair saves corrupted by disease handling matching "herb" inside "catherby"
+            for (variable in listOf("farming_fruit_tree_patch_catherby", "farming_veg_patch_catherby_north", "farming_veg_patch_catherby_south", "farming_flower_patch_catherby")) {
+                val value: String? = this[variable]
+                if (value != null && value.startsWith("herb_dead_")) {
+                    clear(variable)
+                }
+            }
             if (!contains("farming_offset_mins")) {
                 set("farming_offset_mins", random.nextInt(0, 30))
             }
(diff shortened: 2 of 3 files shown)


HarleyGilpin:
Added a spawn repair for existing saves: patches stuck at `herb_dead_N` in the four Catherby non-herb varbits are reset to default on login, same approach as the #1213 compost bin repair. The original crop was overwritten when the value corrupted, so a weedy patch is the best restore. Covered by a new FruitTreePatchTest case.

GregHib (approved):

HarleyGilpin:
More corrupted values turned up in testing: `whiteberry_watered_0` on varbit 729 (`farming_flower_patch_catherby`). Root cause is cross-patch planting — the weeded branch of `FarmingPatchPlant.plantSeed` calls `plant()` without ever checking the item's `farming_patch` type, so a bush seed can be planted in a flower patch. The check in the other branch was also dead for tree seeds (compared "fruit tree" against "fruit_tree").

a048fca adds:
- a shared seed-to-patch check used by both branches, derived from the variable name (`veg` maps to `allotment`)
- generalises the login repair to clear any patch value missing from its varbit map (subsumes the Catherby `herb_dead` list, and heals the whiteberry flower patch); stuck values can never be sent or grow so there's nothing to preserve
- migrates boolean `introducing_explorer_jack_task` values from the old unlock command to `"completed"`, writing the map directly so task completion doesn't fire twice

HarleyGilpin:
Turns out varp 281 wasn't a stale dev save after all — `TaskSystem` writes `"incomplete"` every time the task system opens before the tutorial's done, and the map only has `unstarted/started/completed`. fad2940 writes `"started"` instead and migrates saves holding the old value on login. That accounts for the last live source of the warnings in #1218; nothing reads `"incomplete"` anywhere.

PR #1226, opened 2026-08-28
Fix assistance not applying to ancient effigy investigations

HarleyGilpin:
Follow-up to #1221. Manual testing showed the level check still blocked an assisted player even when the assistant had the required level. Two separate bugs:

**Skill toggles defaulted to off.** `Assistance.assistant()` only accepts an assistant whose `assist_toggle_<skill>` is on, but every toggle started disabled, so assistance did nothing until the assistant manually clicked each skill. Authentic behaviour is the opposite: "All of the skills are shared by default, though the assisting player can click on any skill to switch it off" ([Assist System](https://runescape.wiki/w/Assist_System), same text in the June 2011 revision). `setupAssistant` now enables every skill the assistant is high enough to share (existing `canAssist` check, with the existing message when some are skipped), and clicking a skill still toggles it off.

**The 30k cap ended assistance mid-session.** Per the wiki, capped assistants "can continue to assist without receiving any extra experience", so the redirects now stay in place: the effigy keeps advancing, overflow experience is lost, and the 24-hour timeout still starts once at the cap (the `exceededMaximum` guard stops it restarting on later drops). New assist requests are still refused while capped.

Tests: effigy assist setup no longer toggles a skill, so it covers the default-share path; added per-stage assist tests for nourished/sated/gorged, a full four-stage chain with one capped assistant (exactly +30,000 to the assistant, lamp produced), a test that a switched-off skill blocks investigation, and RequestAssist tests for default sharing and for lower-level skills being skipped. `Grant exp to assistant` now sets levels before the session starts, since redirects begin at setup and `Experience.set` ignores blocked skills.

### DIFF
--- a/game/src/main/kotlin/content/social/assist/RequestAssist.kt
+++ b/game/src/main/kotlin/content/social/assist/RequestAssist.kt
@@ -83,6 +83,7 @@ class RequestAssist : Script {
                 val maxed = exceededMaximum(gained)
                 player.exp(skill, exp)
                 player["total_xp_earned"] = gained.toInt()
+                // Capped assistants can keep assisting but no longer gain any experience
                 if (maxed) {
                     player.interfaces.sendText(
                         "assist_xp",
@@ -93,7 +94,6 @@ class RequestAssist : Script {
                         """,
                     )
                     player.start("assist_timeout", TimeUnit.HOURS.toSeconds(24).toInt())
-                    stopRedirectingAllExp(this)
                 }
             }
         }
@@ -141,7 +141,7 @@ class RequestAssist : Script {
             sendText("assist_xp", "description", "The Assist System is available for you to use.")
             sendText("assist_xp", "title", "Assist System XP Display - You are assisting ${assisted.name}")
         }
-        applyExistingSkillRedirects(player, assisted)
+        shareSkillsByDefault(player, assisted)
         setAssistAreaStatus(player, true)
         player.sendVariable("total_xp_earned")
         player.anim("assist")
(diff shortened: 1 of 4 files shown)


game/src/main/kotlin/content/social/assist/RequestAssist.kt
```diff
@@ -93,7 +94,6 @@ class RequestAssist : Script {
                         """,
                     )
                     player.start("assist_timeout", TimeUnit.HOURS.toSeconds(24).toInt())
-                    stopRedirectingAllExp(this)
```
GregHib, line 96: But if we don't stop it then all future exp for that player will be lost?

HarleyGilpin:
Pushed d7c7979: manual testing surfaced "An assist request has been refused. You can assist again in 0 hours." Three problems around the timeout:

- `start("assist_timeout", ...)` used the default game-tick base while `getHoursRemaining` read it with an epoch-seconds base, so the timer always looked expired.
- The remainder was converted with `TimeUnit.MILLISECONDS.toHours` on a value in seconds, hence "0 hours". Now rounds up to whole hours, so it reads "24 hours" right after capping.
- `total_xp_earned` was never reset, so a capped assistant was refused forever. `hasEarnedMaximumExperience` now clears it once the timeout expires, and the blocked-xp handler uses the same check so a capped assistant earns again after 24 hours.

Old tick-based `assist_timeout` values in existing saves read as expired against the epoch base, so already-capped players reset on their next request. Two new tests: refusal message with hours remaining, and cap reset after expiry.

game/src/main/kotlin/content/social/assist/RequestAssist.kt
HarleyGilpin: Yes, while the session stays open any exp the assisted player earns in the shared skills is lost once the assistant is capped. That matches the wiki: assistants "can continue to assist without receiving any extra experience" past the 30k cap, and the assisted player never receives exp for shared skills during a session either way - the cap only changes whether the assistant gets it.

Stopping the redirects at the cap is what broke effigies: `Assistance.assistant()` requires the exp block, so the moment an assistant capped (starved + nourished = 35k), the sated stage refused their level. It also silently handed exp back to the assisted player mid-session, which isn't how the assist system behaves.

The window is small in practice - redirects only last while the assist window is open and both players are within 20 tiles, and `cancelAssist` still restores everything the moment the session ends. The interface text also tells the assistant they've capped, so either side can close it if they'd rather train normally.

PR #1233, opened 2026-08-30
Fix evil twin and pheasant random event bugs

### LINKED ISSUE #1227: Pheasant random event bugs
* Pheasants should be insta 1 hit kill (no chance to miss)
* Defend anim is incorrect

HarleyGilpin:
Closes #1227

## Evil Twin

With `world.npcs.randomWalk` on, the npc tick re-applies `Wander` over `EmptyMode` every tick, so a grabbed suspect strolled around mid-lift and wandered back out of the jail while the face-north flag was applied. WorldTest forces `randomWalk` off, which is why no test caught it. The grab, the drop at the jail and the re-entry spawn now use `PauseMode`, and the twin turns north on the same tick as the jailed animation, a tick after `clearTransform`, so the face update can't be lost to the transform or the walk's leftover facing.

New test runs the grab with `randomWalk` enabled and asserts the twin ends on the jail tile facing north.

## Freaky Forester (#1227)

* Pheasants die to a single hit from any attack style, even when the accuracy roll misses. Added an `instant_death` npc definition flag, applied in `Target.damageLimitModifiers` alongside `damage_cap` and `immune_death`.
* The defend anim was wrong because pheasants used the chicken combat definition, and the chicken anims (5387-5389) belong to the reworked chicken rig the pheasant model never got. Pheasants now have their own combat definition with the old bird rig anims (2371-2373) and pheasant sounds (690-692).

New test forces every accuracy roll to fail and asserts one swing kills the pheasant and drops the raw bird; it fails without the fix.


data/activity/event/random/freaky_forester/freaky_forester.combat.toml
```diff
@@ -0,0 +1,12 @@
...
+defend_anim = "pheasant_defend"
+defend_sound = "pheasant_defend"
+death_anim = "pheasant_death"
+death_sound = "pheasant_death"
+
+[pheasant.melee]
+range = 1
+anim = "pheasant_attack"
+target_sound = "pheasant_attack"
+target_hit = { offense = "stab", max = 1 }
```
GregHib:
```suggestion
target_hit = { offense = "stab", max = 0 }
```

Presumably pheasants aren't able to deal damage either

game/src/main/kotlin/content/entity/combat/Target.kt
```diff
@@ -186,6 +186,8 @@ object Target {
             0
         }
         is NPC if target.id == "harpie_bug_swarm" && source is Player && source.equipped(EquipSlot.Shield).id != "lit_bug_lantern" -> 0
+        // Any attack kills outright, even a missed roll (e.g. random event pheasants).
+        is NPC if target.def["instant_death", false] -> target.levels.get(Skill.Constitution)
```
GregHib: `instant_death` seems a bit pre-emptive as I don't know of any other npcs which'll ever need this so let's just do `target.id.startsWith("pheasant")` for now

GregHib (changes requested):

Ebp90:
On my end, none of the random events play the poof spotgfx

data/activity/event/random/freaky_forester/freaky_forester.combat.toml
HarleyGilpin: Applied. Also dropped the leftover `max_hit_melee = 10` from the npc definition — `Damage.maximum` reads `max_hit_$type` off the npc def first, so the combat definition's `max = 0` would have been ignored while it was there. 3ed58385ff028fdc7f00e8010d89d98374bdd588

game/src/main/kotlin/content/entity/combat/Target.kt
HarleyGilpin: Switched to `target.id.startsWith("pheasant")` and removed the `instant_death` param. 3ed58385ff028fdc7f00e8010d89d98374bdd588

HarleyGilpin:
Closed - please see new PR request https://github.com/GregHib/void/pull/1245

PR #1235, opened 2026-08-30
Create a check for EssenceMine.kt to prevent a minor softlock

PrettyPrincessPotato:
image
Added a simple check to prevent the player from teleporting to two specific tiles seen in the screenshot.


GregHib:
Thanks, rather than adding some manual checks for these specific tiles we can instead modify the area polygon to avoid that location

PR #1237, opened 2026-08-30
Add Isafdar navigation obstacles and Lletya content

HarleyGilpin:
Adds the interactable objects needed to navigate Isafdar and the missing Lletya content, plus a crystal bow combat fix found while testing.

## Isafdar navigation
- **Dense forest passages** (3937-3939, 3998/3999) - squeeze through with rotation-aware movement; the two Elf Camp passages require 56 Agility (data-driven `level` param). Players line up with the tile nearest the passage centre before squeezing through.
- **Tripwire (3921) / sticks trap (3922)** - step-over with an agility success roll; failure deals damage with authentic messages.
- **Leaf traps (3923-3925)** - jump across on click; failing the roll or walking onto the trap tile drops the player into the pit below (spikes damage), with the protruding rocks (3927) to climb back out. All four trap sites wired to their pits.
- **Log balances (3931/3932/3933)** - 45 Agility, 7.5 xp. Approach is interruptible by walking away; the crossing itself is done at walk speed with the rope balance render emote.
- **Elven Overpass rocks (9296/9297)** - three cliffside shortcuts at 59/68/85 Agility.
- **Lletya tree entrance (8742)** - gated on Roving Elves.
- **Arandar heavy gates (3944/3945)**, **Underground Pass door pass-through (4006/36000)** and **Poison Waste caves/sewer ladders** (until the full dungeon content exists).

## Lletya
- **Teleport crystal** - Activate teleports to Lletya using the existing charge degrade chain, with degradation messages.
- **Eluned** - re-enchants uncharged crystals for 750 coins, decreasing by 150 per recharge to a floor of 150 (persisted per player).
- **Islwyn** - sells new crystal bows/shields for 900,000 and attunes crystal seeds for 900,000 decreasing by 180,000 to a floor of 180,000 (persisted per player).
- **Shopkeeper dialogues** (Eudav, Oronwen, Dalldav, Gethin) opening their existing shops, and flavour dialogue for Goreu, Ysgawyn, Mawrth, Arvel, Kelyn, Eoin, Iona and Gwir.
- Elf musician (8723) no longer turns to face players (`interacts = false`), staying locked facing south.

## Fixes
- **Crystal bow dropped itself as ammo**: `new_crystal_bow` didn't match the `startsWith("crystal_bow")` check in `ammo/Ammo.kt`, so firing it removed the bow from the weapon slot and dropped it at the target like a thrown weapon. Now maps to the same no-ammo handling as the rest of the chain.
- **`objectApproach` supports `arrive = false`** (mirroring `objectOperate`): approach handlers can now run without waiting for the already-queued route to finish. Needed because the log balance's cache reach tile is south of the log - without this the player always detoured a tile south before crossing.

## Notes
- Trap fail damages and success rates are placeholder values flagged with TODOs; exact 2011 numbers are unverified.
- Which pit belongs to the west/centre/north leaf traps is a best guess (the south trap's pit is verified); flagged in code.
- Quest gates use `questCompleted("roving_elves")`, relying on `quests.requirements.skipMissing` until the elf quests are implemented.

## Testing
- `IsafdarTest` (17 tests): every obstacle type, level gates, trap failure damage, pit fall/climb-out, log pathing regression (7 start positions asserting no south detour), approach interruption.
- `LletyaTest` (6 tests): teleport crystal charge chain, Eluned/Islwyn price ladders and persistence, shopkeeper dialogue opening shops.
- `CrystalBowTest`: bow stays equipped and degrades correctly when fired.
- Full `:game:` and `:engine:` suites pass.


Ebp90:
I can convert Regicide, Roving Elves and MEP1 to compliment this at some point.

GregHib (changes requested):
Dense forests:
- anims don't seem to be correct
- if you stand to the side of one and enter you can clip yourself into the forest on the other side
- Anim finishes too early and force walk slides the last tick

Sticks:
- Can move diagonal
- Anim finishing too early and slides at end
- You can step onto them without setting them off

Tripwire:
- A little janky with teleports/anims but generally okay

Leaf traps:
- Also diagonal movement issue
- Doesn't trigger when walking over without interaction
- Also slide at end of movement

Log balance:
- Run orb disables when walking across (see other log balances)

HarleyGilpin:
Addressed the review in bf7c86f39, using a packet capture from OSRS as reference:

**Dense forests**: animations were wrong for most ids: 3937/3999 use 839 (climb-over), 3938/3998 use 749 (double squeeze, cleared after the move), 3939 uses 1237 (side squeeze), each with their forest sound. The exact move is now 94cs with a 30cs start delay which removes the early anim finish/slide. The line-up walk is now collision-aware (`walkToDelay` instead of a forced walk-over) and aborts if the lane tile isn't reached, so entering from the side can't clip through anymore.

**Tripwire**: reworked to the captured shape: walk onto the near wire tile, single-tile exact move over the wire with anim 1236, then step off. Failing now animates the flanking arrow launchers (3920) and fires arrow projectiles at the player.

**Sticks**: crossing walks straight along the trap at walk speed (lined up first, so no diagonals). Failing springs the trap (obj anim + stumble-back 846) and flings the player off in the fixed direction given by the trap's rotation. Walking onto the trap without interacting now sets it off the same way, with a regression test.

**Leaf traps**: trigger areas now cover the full 3x3 leaf patches, so walking or cutting a diagonal over any leaf tile drops the player into the pit (stumble 764 → getup 534 + \"Ouch!\"), with a test. Jumps line up with the trap centre first and use the captured timings, which also fixes the end slide. Scripted crossings suppress the walk-on trigger via a var.

**Log balance**: the run setting is no longer touched; crossing relies on force-walk like the other log balances so the run orb stays lit, and the balance sound plays each step.

One deviation: in our map data the tripwire midline carries wall flags, so walking through the wire without interacting is collision-blocked, the capture's walk-through arrow volley can't trigger and I left that walk-on handler out rather than firing it from the approach tile.

HarleyGilpin:
Follow-up in e4b63956f: the walk-through deviation from the previous comment is resolved, the tripwire tiles are now tagged as `border` areas, so walking at the wire crosses it the same way the border guard gates work (forced walk through the wall-flagged midline), with the arrow volley, damage and poison firing on the way through. Interacted step-overs still suppress the volley and roll as before.

HarleyGilpin:
@GregHib - All adjustments have been made.

game/src/main/kotlin/content/area/tirannwn/isafdar/IsafdarTraps.kt
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }
+        sound("tripwire_with_arrows", delay = 20)
+    }
+
+    /**
+     * The near trap tile, far trap tile, tile beyond the far side and crossing
+     * direction for stepping over [target] from the player's side of it.
+     */
+    fun Player.crossing(target: GameObject): Crossing {
```
GregHib, line 172: The stick trap at 2200, 3169 overshoots the player into the tree rather than walk the player to the south instead, might be easier to just pass a target tile and store them in a tables.toml

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
```diff
@@ -37,6 +40,16 @@ class LogBalance : Script {
             message("... and make it safely to the other side.", ChatType.Filter)
         }
 
+        objectApproach("Cross", "isafdar_log_balance,isafdar_log_balance_2,arandar_log_balance", arrive = false) { (target) ->
+            approachRange(10)
+            steps.clear()
+            crossLog(target)
+        }
```
GregHib: I don't think approach is needed for these (when I tested it gave me a message saying can't reach that)

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
```diff
@@ -76,4 +89,39 @@ class LogBalance : Script {
...
+        val to = if (from == start) end else start
+        // Approach the log at normal speed; walking away cancels
+        walkTo(from)
+        var count = 0
+        while (tile != from && count++ < 50) {
+            pause(1)
+        }
+        if (tile != from) {
+            return
+        }
```
GregHib: walkOverDelay should be fine for this

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
```diff
@@ -76,4 +89,39 @@ class LogBalance : Script {
...
+        message("You walk carefully across the slippery log...", ChatType.Filter)
+        delay()
+        renderEmote("rope_balance")
+        walkTo(to, noCollision = true, forceWalk = true)
+        var steps = 0
+        while (tile != to && steps++ < 10) {
+            sound("log_balance")
```
GregHib: Looks like the rope_balance render emote already has sounds so this must be osrs specific

GregHib (changes requested):

game/src/main/kotlin/content/area/tirannwn/isafdar/IsafdarTraps.kt
HarleyGilpin: Moved the landing tiles into `isafdar.tables.toml` (`sticks_traps`, keyed by trap tile with a destination per direction) in a2a5df7e8, the (2200,3169) trap now walks the player out to (2202,3168), south past the tree, and the fling uses the same table. Covers all six traps with collision-checked tiles.

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
HarleyGilpin: The "can't reach that" comes from clicking the far log object, which the pathfinder can't reach, the original game gives the same message there. The approach handler is still needed for the near log though: without it the interaction pathfinds to the melee reach tile diagonally behind the log (2258,3249) and walks the whole route before operating, so the player dips a tile south before crossing. The `Crossing the log never paths south around it` test reproduces it if the approach is removed. Happy to rework it if you'd prefer a different shape for skipping the arrival delay.

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
HarleyGilpin: Done in a2a5df7e8 — the approach walk is a plain `walkOverDelay` now.

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
HarleyGilpin: Removed the manual sound loop and the sound definition in a2a5df7e8, relying on the render emote's own sounds.

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
HarleyGilpin: walkOverDelay's delay is non-cancellable, and it sets the "delay" var which makes the Walk instruction ignore input, so the player can't cancel by walking away while still approaching the log (they'd be dragged the rest of the way and across). The pause loop keeps the approach interruptible; there's a test covering it (`Walking away while approaching the log cancels the crossing`). The crossing itself still uses walkOverDelay.

GregHib: Yeah but as the approach is removed it shouldn't be an issue as it won't activate until right next to the log (and you can't cancel once on the log)

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
GregHib: > without it the interaction pathfinds to the melee reach tile diagonally behind the log (2258,3249) and walks the whole route before operating

This doesn't seem to be the case, without an approach it walks to 2259, 3249 or 2258, 3250

HarleyGilpin: You're right, retested and the reach tile lands beside the log, so the approach handler and the `arrive` parameter engine addition are both removed in 5846c3486. The far-end "can't reach that" matches the original game, and there's a test covering it now.

game/src/main/kotlin/content/skill/agility/shortcut/LogBalance.kt
HarleyGilpin: Done in 5846c3486, approach removed and the whole thing is walkOverDelay now. One engine fix fell out of it: Movement.calculate was queueing steps with only the equipment-based forceWalk check and discarding the noRun bit that walkTo packs into the target step, so walkOverDelay(forceWalk = true) crossings actually ran at run speed (this affected the other log balances too). It now honours the packed bit.

engine/src/main/kotlin/world/gregs/voidps/engine/entity/character/mode/move/Movement.kt
```diff
@@ -50,9 +50,9 @@ open class Movement(
         val pathfinds = character is Player || (character is NPC && character["owner_index", -1] != -1)
         if (pathfinds && !tile.noCollision) {
             val route = pathFinder.findPath(character, strategy, shape)
-            character.steps.queueRoute(route, tile, tile.noCollision, noRun = strategy.forceWalk(character))
+            character.steps.queueRoute(route, tile, tile.noCollision, noRun = tile.noRun || strategy.forceWalk(character))
         } else if (tile != Tile.EMPTY) {
-            character.steps.queueStep(tile, tile.noCollision, noRun = strategy.forceWalk(character))
+            character.steps.queueStep(tile, tile.noCollision, noRun = tile.noRun || strategy.forceWalk(character))
         }
```
GregHib, line 56: Seems redundant, noRun isn't set on strategy.destination (which are tiles not steps) and if you were, you can do it with strategy.forceWalk

HarleyGilpin: Moved the check into `TileTargetStrategy.forceWalk` as suggested in f5820e0a3 so Movement is untouched.

game/src/main/kotlin/content/area/tirannwn/lletya/LletyaTreeEntrance.kt
```diff
@@ -0,0 +1,25 @@
...
+            if (!questCompleted("roving_elves")) {
+                statement("The trees are too dense for you to find a way through.")
+                return@objectOperate
+            }
+            val direction = if (tile.x < target.tile.x) Direction.EAST else Direction.WEST
+            val dest = Tile(target.tile.x + direction.delta.x * 2, target.tile.y, tile.level)
+            face(direction)
+            delay()
+            anim("dense_forest_squeeze")
+            exactMoveDelay(dest, delay = 90, direction = direction)
```
GregHib: This isn't the correct anim, the objects are animated and open up and then the player walkOverDelay's to the other side

game/src/main/kotlin/content/area/tirannwn/poison_waste/PoisonWasteDungeon.kt
```diff
@@ -0,0 +1,45 @@
...
+            anim("climb_down")
+            delay(2)
+            val dest = if (target.tile.equals(2041, 4189) || target.tile.equals(2041, 4172)) {
+                target.tile.add(-1, 0, -1)
+            } else {
+                target.tile.add(0, -1, -1)
+            }
+            tele(dest)
+        }
+    }
```
GregHib: These can all be .teles.toml

game/src/main/kotlin/content/area/tirannwn/isafdar/LeafTraps.kt
```diff
@@ -0,0 +1,109 @@
...
+            this["crossing_trap"] = true
+            if (tile != start) {
+                // Line up with the trap before jumping
+                walkToDelay(start, forceWalk = true)
+                if (tile != start) {
+                    clear("crossing_trap")
+                    return@objectOperate
+                }
+            }
+            if (Level.success(levels.get(Skill.Agility), 128..250)) { // TODO unknown rate
```
GregHib: https://oldschool.runescape.wiki/w/Leaves_(trap)
```suggestion
            if (Level.success(levels.get(Skill.Agility), 180..255)) {
```

game/src/main/kotlin/content/area/tirannwn/isafdar/LeafTraps.kt
```diff
@@ -0,0 +1,109 @@
...
+            Site(centre = Tile(2277, 3262), pit = Tile(2354, 9643), exit = Tile(2275, 3262)),
+        )
+
+        private fun fall(player: Player, site: Site) {
+            player.message("You fall through and onto some spikes.")
+            player.tele(site.pit)
+            player.anim("pit_getup")
+            player.say("Ouch!")
+            player.sound("fall_land")
+            player.damage(random.nextInt(20, 41)) // TODO unknown damage
```
GregHib:
```suggestion
            player.damage(random.nextInt(20, 181)) // TODO unknown min damage
```

game/src/main/kotlin/content/area/tirannwn/isafdar/DenseForest.kt
```diff
@@ -0,0 +1,64 @@
...
+                    sound("forest_doublesqueeze", delay = 30)
+                }
+                else -> {
+                    anim("dense_forest_squeeze", delay = 30)
+                    sound("forest_sidesqueeze", delay = 40)
+                }
+            }
+            exactMoveDelay(dest, delay = 94, direction = direction, startDelay = 30)
+            // Interacting faced the object, which is now behind; keep facing the way travelled
+            face(direction)
```
GregHib, line 61: Automatic due to how interactions work
```suggestion
```

game/src/main/kotlin/content/area/tirannwn/isafdar/Eluned.kt
```diff
@@ -0,0 +1,54 @@
...
+        val price = rechargePrice()
+        npc<Happy>("Very well. I'll recharge your teleportation crystal for $price gold. What do you say?")
+        choice {
+            option<Neutral>("Recharge a crystal.") {
+                if (!inventory.contains("coins", price)) {
+                    player<Sad>("Actually, I don't have enough coins.")
+                    return@option
+                }
+                if (inventory.remove("coins", price) && inventory.remove("crystal_teleport_seed_uncharged")) {
+                    inventory.add("crystal_teleport_seed_4")
```
GregHib: Transaction

game/src/main/kotlin/content/area/tirannwn/isafdar/IsafdarTraps.kt
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+
+        objectOperate("Step-over", "tripwire") { (target) ->
+            val (entry, exit, out, direction) = crossing(target)
+            this["crossing_trap"] = true
+            walkOverDelay(entry)
+            anim("tripwire_step_over")
+            exactMoveDelay(exit, delay = 30, direction = direction)
+            if (Level.success(levels.get(Skill.Agility), 128..250)) { // TODO unknown rate
```
GregHib: https://oldschool.runescape.wiki/w/Tripwire
```suggestion
            if (Level.success(levels.get(Skill.Agility), 30..155)) {
```

game/src/main/kotlin/content/area/tirannwn/isafdar/IsafdarTraps.kt
```diff
@@ -0,0 +1,201 @@
...
+            walkOverDelay(out)
+            clear("crossing_trap")
+        }
+
+        objectOperate("Pass", "isafdar_sticks_trap") { (target) ->
+            val (entry, _, out, direction) = crossing(target)
+            this["crossing_trap"] = true
+            walkToDelay(entry.add(direction.inverse()), forceWalk = true)
+            walkOverDelay(entry)
+            if (Level.success(levels.get(Skill.Agility), 31..156)) {
```
GregHib: ? https://oldschool.runescape.wiki/w/Sticks_(trap)
```suggestion
            if (Level.success(levels.get(Skill.Agility), 30..155)) {
```

game/src/main/kotlin/content/area/tirannwn/isafdar/ElvenOverpass.kt
```diff
@@ -0,0 +1,32 @@
...
+import world.gregs.voidps.engine.entity.character.player.chat.noInterest
+import world.gregs.voidps.engine.entity.character.player.chat.obstacle
+import world.gregs.voidps.engine.entity.character.player.skill.Skill
+import world.gregs.voidps.engine.entity.character.player.skill.level.Level.has
+import world.gregs.voidps.type.Tile
+
+class ElvenOverpass : Script {
+
+    init {
+        objectOperate("Climb", "elven_overpass_rocks_up,elven_overpass_rocks_down") { (target) ->
```
GregHib: The timing/anims on these are off.

Dump:
[elven-overpass.txt](https://github.com/user-attachments/files/31954356/elven-overpass.txt)

GregHib (changes requested):

game/src/main/kotlin/content/area/tirannwn/lletya/LletyaTreeEntrance.kt
HarleyGilpin: Reworked in 29a03c7b2, both tree gate objects play the open animation with the door sounds.

game/src/main/kotlin/content/area/tirannwn/poison_waste/PoisonWasteDungeon.kt
HarleyGilpin: Done in 29a03c7b2, script deleted, entrances, exits and all three ladders are entries in poison_waste.teles.toml.

game/src/main/kotlin/content/area/tirannwn/isafdar/LeafTraps.kt
HarleyGilpin: Applied in 29a03c7b2.

game/src/main/kotlin/content/area/tirannwn/isafdar/LeafTraps.kt
HarleyGilpin: Applied in 29a03c7b2.

game/src/main/kotlin/content/area/tirannwn/isafdar/DenseForest.kt
HarleyGilpin: Player faces behind them unless i set the face direction.

game/src/main/kotlin/content/area/tirannwn/isafdar/Eluned.kt
HarleyGilpin: Done in 29a03c7b2, the recharge is a single transaction (coins + uncharged seed out, charged seed in) with the deficient case falling back to the apology line.

game/src/main/kotlin/content/area/tirannwn/isafdar/IsafdarTraps.kt
HarleyGilpin: Applied in 29a03c7b2.

game/src/main/kotlin/content/area/tirannwn/isafdar/IsafdarTraps.kt
HarleyGilpin: Applied in 29a03c7b2 (and matched the sticks operate roll to it).

game/src/main/kotlin/content/area/tirannwn/isafdar/ElvenOverpass.kt
HarleyGilpin: Reworked from the dump in 29a03c7b2, climbing down is the exact move facing the cliff (delay 120/90 per site) with human_climbing_down and the climbing_loop sound, climbing up walks the cliff path tile by tile under the climbing render emote, and the level gate uses the dump's message. One note: the dump's objects are 16514/16515 but our cache has 9296/9297 at those exact tiles (16514/16515 don't exist in 634 here), so the ids stay, renamed the defs top/bottom to match. The dump's overpass gates (3944/3945) are also reworked to swing open to their open variants, walk the player through and close five ticks later.

game/src/main/kotlin/content/area/tirannwn/isafdar/DenseForest.kt
```diff
@@ -0,0 +1,64 @@
...
+
+class DenseForest : Script {
+
+    init {
+        objectOperate("Enter", "dense_forest,dense_forest_2,dense_forest_3,dense_forest_hard,dense_forest_hard_2") { (target) ->
+            val level = target.def["level", 0]
+            if (level > 0 && !has(Skill.Agility, level)) {
+                obstacle(level)
+                return@objectOperate
+            }
```
GregHib: > you do not need 56 Agility to pass back through dense 
https://oldschool.runescape.wiki/w/Dense_forest

GregHib (approved):

PR #1238, opened 2026-08-30
Add scene editor persist commands (place/remove/flush + JS5)

MaatheusGois:
# Server-backed scene editor persistence and tooling

Screenshot 2026-09-06 at 13 05 00

## Summary

This PR provides the server-side persistence and command hooks used by the void-client scene editor.

- Adds admin commands for authoritative scene updates:
  - `scene_place <object-id> <x> <y> <plane> [rotation] [shape]`
  - `scene_remove <object-id> <x> <y> <plane> [rotation] [shape]`
  - `scene_flush`
  - `scene_status`
- Persists object placements and removals under `data/area/scene/editor.obj-spawns.toml`.
- Applies changes to live `GameObjects`, including removals of stock map objects, and preserves the actual object shape/rotation in removal tombstones.
- Adds server-backed NPC scene support:
  - `scene_npc_spawn <npc-id> <x> <y> <plane>`
  - `scene_npc_remove <npc-id> <x> <y> <plane>`
  - NPC placements persist in `data/area/scene/editor.npc-spawns.toml` and use the normal authoritative NPC update, interaction, animation, and combat systems.
- Adds item actions used by the editor:
  - `scene_item_drop <item-id> <x> <y> <plane> [amount]`
  - `scene_item_bag <item-id> [amount]`
- Keeps deployment-specific scene data out of the repository with `/data/area/scene/` ignored.
- Keeps runtime cache rewriting out of the server. `CacheDelegate`, Koin cache registration, and runtime JS5 writes are not required by scene-editor commands.
- Provides the explicit `tools:updateObjectSpawns` workflow for intentionally converting persisted object-spawn data into cache map updates.

## Persistence behavior

- Object/NPC place and remove commands update the live server state and persist immediately.
- `scene_flush` reloads persisted scene-editor object changes for clients that need an explicit refresh.
- NPC removal persists a tombstone at the original spawn tile so wandering NPCs do not reappear after a restart.
- Scene files are deployment-local and may differ between server installations.

## Verification

- [x] `./gradlew :game:compileKotlin`
- [x] `./gradlew :game:shadowJar`
- [x] `./gradlew :game:test`
- [ ] Run the freshly built server on Java 21+ and verify object placement/removal through the client editor.
- [ ] Restart the server and verify removed objects and NPCs do not return.
- [ ] Spawn an NPC through the catalog and verify normal interaction, menus, animation, and combat.
- [ ] Drop an item and add an item to the player's bag through the editor actions.

The generated files under `data/area/scene/` are intentionally not included in this PR.


GregHib:
A bit of separation of concerns would be good here, a command which updates an `.obj-spawns.toml` and auto refreshes seems reasonable and the cache updating which takes a toml and writes it to maps in the cache would be better suited as a script in `/tools`.

Reasoning for this is the cache is removed after startup to reduce unnecessary memory overhead as so is intentionally left out of koin to avoid runtime dependencies, adding it and the ~100MB+ of ram usage back in for what is a dev convenience isn't worth it imo.

MaatheusGois:
@GregHib done!

GregHib:
Thanks, this is quite a different beast of a PR now.

Void's aim is to accurately emulate rs both in terms of content and behaviour but also in-a-sense code as well, trying to keep to the core minimal systems and while there is some give for functionality that makes life easier (moderation, analytics etc..) a custom editor and systems to support it, while very cool, steps over that line too much, I'm all for people using and customising it to their liking but that's what forks are for

PR #1239, opened 2026-08-31
Add remaining Hunter methods: box traps, deadfalls, magic boxes, falconry and kebbit tracking

HarleyGilpin:
Finishes the remaining Hunter training methods (bird snares, nets, pitfalls, butterflies and implings were already in).

## Box traps
Lay/Check/Dismantle/Investigate on `box_trap`, sharing the bird snare lifecycle. Catches grey/carnivorous chinchompas, ferret (Eagles' Peak-gated), pawya and grenwall (bait-gated with papaya/raw pawya meat).

## Deadfall traps
Set-trap on a boulder with knife + logs (level 23). Catches wild, barb-tailed, prickly and sabre-toothed kebbits. Logs consumed on trigger, returned on dismantle.

## Magic box / imp catching
Level 71 magic box catches imps only. Imp-in-a-box supports Talk-to dialogue and Bank (authentic Imp Release Negotiation Form, interface 478).

## Falconry
Matthias lends a gyr falcon (Hunter 43, both hands free). Catches spotted/dark/dashing kebbits from range; falconer's gloves occupy the weapon slot for the loan's duration.

## Kebbit tracking
Adds polar, common, feldip weasel, razor-backed and desert devil trails. Trail data is TOML-driven; track-decal varbits were included. Trails are session-state, cleared on logout.

## Unfinished
Rabbit snaring is still being worked on.

## Tests
`BoxTrapTest`, `DeadfallTest`, `MagicBoxTest`, `FalconryTest`, `KebbitTrackingTest` — 31 new WorldTest cases; full `:game:test` green.


game/src/main/kotlin/content/skill/hunter/BoxTrap.kt
```diff
@@ -0,0 +1,183 @@
...
+
+        objectOperate("Investigate", "box_trap") { (target) ->
+            val id = Tables.npc("traps.box_trap.npc")
+            val npc = NPCs.find(target.tile, id)
+            Traps.investigate(this, npc)
+        }
+
+        itemOnObjectOperate("*", "box_trap") {
+            when {
+                it.item.id == "unlit_torch" -> message("I should light the torch before using it to smoke the trap.")
```
GregHib, line 61: Smoking box traps gives "Nothing interesting message"

game/src/main/kotlin/content/skill/hunter/Traps.kt
```diff
@@ -1,18 +1,97 @@
...
+        }
+        if (floorItem != null) {
+            if (!FloorItems.remove(floorItem)) {
+                return
+            }
+        } else if (!player.inventory.remove(trapId)) {
+            return
+        }
+        player.inc("trap_count")
+        NPCs.add(Tables.npc("traps.$trapId.npc"), player.tile, ticks = 100, owner = player)
```
GregHib, line 63: I think chinchompa's should be able to walk over boxes but I'm not sure what about the collision allows them to do this? Maybe it's an NPC check rather than a collision check and the box doesn't have collision on spawn?

game/src/main/kotlin/content/skill/hunter/Falconry.kt
```diff
@@ -0,0 +1,189 @@
...
+            if (lifecycle != 0) {
+                return@npcDespawn
+            }
+            if (player.equipped(EquipSlot.Weapon).id == "falconers_glove") {
+                player.equipment.replace("falconers_glove", "falconers_glove_2")
+                player.message("Your falcon gives up on its catch and returns to your glove.")
+            }
+        }
+
+        exited("piscatoris_falconry_area") {
```
GregHib, line 90: Looks like the area needs fixing as if you walk just a bit north inside the enclosure the bird disappears

game/src/main/kotlin/content/skill/hunter/Falconry.kt
```diff
@@ -0,0 +1,189 @@
...
+                option<Neutral>("No thanks, I'm fine.")
+            }
+        }
+
+        npcOperate("Falconry", "matthias") {
+            startFalconry()
+        }
+
+        npcApproach("Catch", "spotted_kebbit,dark_kebbit,dashing_kebbit") { (target) ->
+            approachRange(6)
```
GregHib: Might need to clear movement after this, currently it runs right up to the kebbit after

game/src/main/kotlin/content/skill/hunter/Falconry.kt
```diff
@@ -0,0 +1,189 @@
...
+            catch(target)
+        }
+
+        npcOperate("Retrieve", "spotted_kebbit_caught,dark_kebbit_caught,dashing_kebbit_caught") { (target) ->
+            retrieve(target)
```
GregHib, line 75: The caught bird likes to walk around? Not sure if this is correct

game/src/main/kotlin/content/skill/hunter/Falconry.kt
```diff
@@ -0,0 +1,189 @@
...
+                return@npcDespawn
+            }
+            if (player.equipped(EquipSlot.Weapon).id == "falconers_glove") {
+                player.equipment.replace("falconers_glove", "falconers_glove_2")
+                player.message("Your falcon gives up on its catch and returns to your glove.")
+            }
+        }
+
+        exited("piscatoris_falconry_area") {
+            if (hasFalcon()) {
```
GregHib, line 91: Leaving the area doesn't remove a bird on the floor which has a catch

game/src/main/kotlin/content/skill/hunter/DeadfallTrap.kt
```diff
@@ -0,0 +1,208 @@
...
+import world.gregs.voidps.engine.inv.remove
+
+class DeadfallTrap : Script {
+    init {
+        objectOperate("Set-trap", "boulder_trap") { (target) ->
+            setTrap(target)
+        }
+
+        objectOperate("Dismantle", "boulder_trap_setup") { (target) ->
+            dismantleTrap(target, null)
```
GregHib, line 38: Dismantle uses snare anim

game/src/main/kotlin/content/skill/hunter/DeadfallTrap.kt
```diff
@@ -0,0 +1,208 @@
...
+import world.gregs.voidps.engine.entity.item.floor.FloorItems
+import world.gregs.voidps.engine.entity.obj.*
+import world.gregs.voidps.engine.inv.add
+import world.gregs.voidps.engine.inv.inventory
+import world.gregs.voidps.engine.inv.remove
+
+class DeadfallTrap : Script {
+    init {
+        objectOperate("Set-trap", "boulder_trap") { (target) ->
+            setTrap(target)
```
GregHib, line 34: Shouldn't step back after setting a boulder trap I don't think

game/src/main/kotlin/content/skill/hunter/KebbitTracking.kt
```diff
@@ -0,0 +1,240 @@
...
+        message("You discover some tracks nearby.")
+    }
+
+    private fun Player.inspectTrail(target: GameObject) {
+        val trail = trackingTrails[accountName]
+        if (trail == null) {
+            message("You search but find nothing.")
+            return
+        }
+        val step = (trackingSteps[accountName] ?: 0).coerceAtMost(trail.lastIndex)
```
GregHib, line 187: There should only be one trail at a time, I was able to search other areas and uncover a new trail which reset the old one making is near impossible to complete

GregHib (changes requested):
I think chinchompa's should be able to walk over boxes

game/src/main/kotlin/content/skill/hunter/Falconry.kt
HarleyGilpin: I checked it on youtube, that is why there is an arrow indicator so you can find your falcon after it catches they prey.

game/src/main/kotlin/content/skill/hunter/BoxTrap.kt
HarleyGilpin: Torch item id was wrong (torch_lit doesn't exist, the lit item is lit_torch), so it always fell through to noInterest(). Same bug in all four trap files, fixed.

game/src/main/kotlin/content/skill/hunter/Traps.kt
HarleyGilpin: It's the collision check — GameObjects.add() defaults collision=true, which registers pathfinder clip flags. Passed collision=false for ground traps so NPCs can walk onto them.

game/src/main/kotlin/content/skill/hunter/Falconry.kt
HarleyGilpin: The area's north boundary (y=3612) cut straight through Matthias's house instead of following the walls, so the walled room continues past it with no gap until the door. Extended the area to y=3620 to cover the full house.

game/src/main/kotlin/content/skill/hunter/Falconry.kt
HarleyGilpin: approachRange(6) defaults update=true, which skips the step-clearing that stops movement once in range (see Attack.kt's ranged case). Passed update=false so the player stops once in falcon range instead of continuing to melee the kebbit.

game/src/main/kotlin/content/skill/hunter/Falconry.kt
HarleyGilpin: Added a cleanup in the exited() handler that finds and removes any owned *_kebbit_caught NPC when leaving the area.

game/src/main/kotlin/content/skill/hunter/DeadfallTrap.kt
HarleyGilpin: Switched the dismantle anim to lay_trap (kneeling, same as setting it) instead of take_trap, which is the snare pickup animation.

game/src/main/kotlin/content/skill/hunter/DeadfallTrap.kt
HarleyGilpin: Removed it — the player interacts with the boulder from an adjacent tile, they're never standing on it, so stepping away didn't make sense.

game/src/main/kotlin/content/skill/hunter/KebbitTracking.kt
HarleyGilpin: Fixed — inspectBurrow() unconditionally overwrote any trail in progress. It now only lets you restart if you're re-inspecting the same burrow your current trail started from; a different burrow while a trail is active gets a message instead.

HarleyGilpin:
@GregHib - Code review comments addressed and tested.

data/skill/hunter/hunter.tables.toml
GregHib, line 244: Box trapping pawya and grenwall's doesn't seem to work

game/src/test/kotlin/content/skill/hunter/BoxTrapTest.kt
```diff
@@ -0,0 +1,260 @@
...
+        })
+        val player = createPlayer()
+        val start = player.tile
+        player.inventory.add("box_trap")
+        player.levels.set(Skill.Hunter, 60)
+
+        player.itemOption("Lay", "box_trap")
+        tick(3)
+        assertTrue(GameObjects.at(start).any { it.id == "box_trap" })
+        createNPC("chinchompa", player.tile.addY(2))
```
GregHib, line 255: Might be worth adding chinchompa exploding when you hit them too

game/src/main/kotlin/content/skill/hunter/Falconry.kt
```diff
@@ -0,0 +1,190 @@
...
+            return
+        }
+        val success = Level.success(levels.get(Skill.Hunter), row.intRange("chance"))
+        if (!success) {
+            target.shoot("gyr_falcon", this)
+            sound("falcon_return", delay = 20)
+            message("The kebbit is too quick for your falcon.")
+            return
+        }
+        equipment.replace("falconers_glove_2", "falconers_glove")
```
GregHib: This should probably happen before the delay(2)? It looks odd that the bird flies and still is on your hand

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+
+    private suspend fun Player.impDialogue() {
+        player<Quiz>("Hey imp, are you still there?")
+        npc<Neutral>("imp", "Of course I can hear ya, ya great big ape. You know, there's not even enuf space to swing Bob about in 'ere. How about a breather? You know, stretch me pins for a bit?")
+        impOptions()
+    }
+
+    private suspend fun Player.impOptions() {
+        choice {
+            option<Neutral>("No, I'm going to keep you in there.") {
```
GregHib: Duplicate dialogues

```suggestion
            option("No, I'm going to keep you in there.") {
```

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+                player<Neutral>("No, I'm going to keep you in there. I might keep you as a pet.")
+                npc<Neutral>("imp", "Pet!! Nah mate. I fink you'd find dat you'd be my pet!! We is not makin good pets.")
+                player<Quiz>("Really? Why not?")
+                npc<Neutral>("imp", "Coz...errr...")
+                npc<Neutral>("imp", "We bite! Yeah we is biting and...and...er...")
+                npc<Neutral>("imp", "We is fire risk! Yeah dat's it! We be burning down your housey and stealin' all ya shiny gems. Oh, and da beads!! Mmmm, beads.")
+                player<Quiz>("Fire risk? How does that work?")
+                npc<Neutral>("imp", "Is those wizzies. Dey don't like de imps so dey make us go BOOOM!!")
+            }
+            option<Neutral>("It's not that bad.") {
```
GregHib:
```suggestion
            option("It's not that bad.") {
```

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+                player<Neutral>("It's not that bad. You've got four big windows, charming company...er...")
+                npc<Neutral>("imp", "Yeah, we's love tiny, crampt space. It be magical. But I is a busy imp, innit? Dragons needin' ticklin', shiny relics needin' stealin', you know how it goes.")
+                npc<Neutral>("imp", "So, if you's know whas good for ya, you'd be lettin' me go, right?")
+                impOptions()
+            }
+            option<Quiz>("Don't I get three wishes?") {
+                player<Quiz>("Don't I get three wishes?")
```
GregHib:
```suggestion
```

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+    }
+
+    private suspend fun Player.bankOptions() {
+        choice {
+            option<Happy>("Okay, that sounds fair.") {
+                open("imp_box")
+            }
+            option<Quiz>("Surely it should be three items?") {
```
GregHib:
```suggestion
            option("Surely it should be three items?") {
```

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+            return
+        }
+        if (inventory.contains("imp_in_a_box_2")) {
+            inventory.replace("imp_in_a_box_2", "imp_in_a_box_1")
+            interfaces.sendText("imp_box", "text", depositText())
+        } else if (inventory.contains("imp_in_a_box_1")) {
+            inventory.replace("imp_in_a_box_1", "magic_box")
+            close("imp_box")
+            message("The imp takes your items to the bank and escapes from the box.")
+        } else {
```
GregHib: Duplicate logic with line 134, extract a method than returns int or something

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    private fun Player.depositText() = if (inventory.contains("imp_in_a_box_2")) {
+        "Select an item or stack of items to deposit. <br>You can deposit up to 2 items or stacks."
+    } else {
+        "Select an item or stack of items to deposit. <br>You can deposit 1 more item or stack."
+    }
+
+    private fun Player.depositItem(item: Item, slot: Int) {
```
GregHib, line 159: Need to make sure this can't be used in certain places (30+ wilderness, clan wars dangerous etc...)

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
```diff
@@ -0,0 +1,274 @@
...
+import world.gregs.voidps.engine.inv.remove
+import world.gregs.voidps.engine.inv.replace
+
+class MagicBox : Script {
+    init {
+        itemOption("Activate", "magic_box") {
+            layTrap(null)
+        }
+
+        floorItemOperate("Lay") { (item) ->
```
GregHib:
```suggestion
        floorItemOperate("Activate") { (item) ->
```

GregHib (changes requested):

data/skill/hunter/hunter.tables.toml
HarleyGilpin: Root cause: detection range. The trap only noticed creatures already within 2 tiles with line of sight, so with Isafdar's sparse spawns, wide wander radius and sight-blocking jungle a pawya/grenwall almost never triggered it within the trap's 100-tick life (reproduced with a randomWalk WorldTest). Raised the box trap's hunt_range to 8 and switched the mode to line_of_walk so baited traps lure creatures from range like they should, the wandering repro now catches reliably. 441e2470a

game/src/test/kotlin/content/skill/hunter/BoxTrapTest.kt
HarleyGilpin: Added, chinchompas killed in combat now burst with the chinchompa_impact gfx + explosion sound (trap-caught ones excluded), with a test. 441e2470a

game/src/main/kotlin/content/skill/hunter/Falconry.kt
HarleyGilpin: Moved the glove swap to the moment the falcon launches; the distance-abort and failed-catch paths return it to the glove when the bird flies back. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Extracted useCharge() returning the remaining charges; both flows switch on it. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Added a wildernessLevel > 30 guard on all three entry points (Bank option, dialogue accept, use-item). Clan wars doesn't have a dangerous-area definition yet, happy to extend the guard once one exists. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Applied — the floor option is Activate in the cache. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Applied all four suggestions, plus the same fix to the 'nothing I need banking' option which had the identical duplicate. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Applied. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Applied. 441e2470a

game/src/main/kotlin/content/skill/hunter/MagicBox.kt
HarleyGilpin: Applied. 441e2470a

data/skill/hunter/hunter.tables.toml
GregHib: Still not working :(

game/src/main/kotlin/content/skill/hunter/Falconry.kt
```diff
@@ -0,0 +1,192 @@
...
+            message("You need a trained falcon to catch this kebbit.")
+            return
+        }
+        if (!has(Skill.Hunter, row.int("level"), message = true)) {
+            return
+        }
+        face(target)
+        sound("falcon_swoop")
+        equipment.replace("falconers_glove_2", "falconers_glove")
+        shoot("gyr_falcon", target)
```
GregHib: Needs gfx times set in the .gfx.toml

`shoot(falcon_travel_anim, sourceIndex = 0, targetIndex = 0, curve = 0, sizeOffset = 0, startTime = 24, endTime = 64, startHeight = 150, endHeight = 30, sourceIndex = 0, targetIndex = 0) // 922`

[falconry.txt](https://github.com/user-attachments/files/31954985/falconry.txt)

GregHib (changes requested):

data/skill/hunter/hunter.tables.toml
HarleyGilpin: I just tested it in game. Pawya's require using Papaya fruit, and Grenwalls require Pawya meat as bait. Chinchompas don't require any bait item.

HarleyGilpin: Found the real culprit this time — the line-of-sight requirement. Detection at range 8 needs a clear straight sight line from the trap, and Isafdar's vegetation blocks it almost everywhere (reproduced: a wandering pawya 2 tiles away was never detected at the tiles I tested). Baited box traps lure by scent, so I've dropped the visual check from the box_trap hunt mode entirely and added a guard that re-arms the trap and frees the creature if it can't actually path to the trap tile. The blocked-spot repro with randomWalk on now catches, and it's kept as a test. 16f68d7c8

game/src/main/kotlin/content/skill/hunter/Falconry.kt
HarleyGilpin: Thanks for the capture, corrected everything against it.

GregHib (approved):

PR #1240, opened 2026-08-31
Add imp catching with magic boxes

HarleyGilpin:
Builds on #1239 (the first commit here is that PR; only the last commit is new).

Adds level 71 imp catching:
- Activate a magic box (10025) to place it (19223), using the existing `hunting_imptrap_npc` (5101) trap NPC with a new id-filtered `magic_box` hunt mode targeting `imp` only (god wars imps excluded).
- A caught imp plays gfx 932 and the box runs through the 19225 -> 19226 shaking states; Retrieve gives an imp-in-a-box(2) (the magic box itself is consumed) plus 450 xp. Failure leaves `Magic box failed` (19224); Deactivate returns the box.
- Imp-in-a-box: using an item on it sends that item to the bank and uses a charge, 10027 -> 10028 -> empty magic box; it refuses to bank other imp boxes. The authentic item-select interface (478) is left as a follow-up — this is the minimal use-item flow.

Options (`Activate`, `Deactivate`, `Retrieve`, `Investigate`) match the 634 cache. Imp catch/fail anims 5218/5285 are from later-revision references.

## Tests
`MagicBoxTest`: catch, fail, level gate, imp-in-a-box charge flow, imp-box-on-imp-box refusal.


HarleyGilpin:
Consolidated into #1239.

PR #1241, opened 2026-08-31
Add falconry kebbit hunting

HarleyGilpin:
Builds on #1239 and #1240 (only the last commit is new).

Adds Piscatoris falconry:
- Matthias (5092): Talk-to dialogue or the quick `Falconry` option lends a gyr falcon for 500 coins — requires Hunter 43 and both hand slots free. The falconer's gloves (10024 falcon / 10023 empty) go straight into the weapon slot; they can't be equipped manually, removed by hand, or have another weapon/shield equipped over them, and walking out of the falconry area hands the bird back automatically (a new `piscatoris_falconry_area` area).
- Catching: `Catch` on spotted (43) / dark (57) / dashing (69) kebbits launches the falcon as a projectile from range. On success the kebbit is replaced by an owned caught `Gyr Falcon` npc (5094-5096, 60s revert to your glove if ignored); `Retrieve` gives the fur + bones and 104/132/156 xp. New `[falconry]` table holds the rows.
- Caught-falcon npc defs and options match the 634 cache; the projectile reuses the existing `gyr_falcon` gfx (927) — later-revision references disagree on the exact projectile id (918/922/927) so this could use an in-game look.
- The existing `falconry_area_stile` climb-over already handles the entrance.

## Tests
`FalconryTest`: borrow (and level/coin rejections), catch + retrieve with xp, failed catch returns the falcon, leaving the area strips the glove.


HarleyGilpin:
Consolidated into #1239.

PR #1242, opened 2026-08-31
Add polar and common kebbit tracking

HarleyGilpin:
Builds on #1239, #1240 and #1241 (only the last commit is new).

Adds noose wand kebbit tracking:
- Polar kebbits (level 1, 30 xp) in the Rellekka snow area and common kebbits (level 3, 36 xp) east of the Piscatoris falconry pen. Inspecting a burrow/hole generates a random trail of 2-3 linking segments (tunnels chain for free), each revealed by inspecting the plant/tunnel/hollow log at the current segment's trigger tile; the final spot (bush 19428 / snow drift 19435) is Searched then Attacked with a noose wand for the fur + bones + raw beast meat.
- Trail data is TOML-driven: a `[tracking]` table for the two kebbits and a `[trails]` table of segments (varbit, tunnel/inverted flags, start/end/trigger tiles, starting burrow); segment linking, inverse-direction trails, and no-varbit-reuse generation happen at runtime.
- Track visuals come from the client's own varbit-transformed decal objects — the 19 trail varbits (2974-2983, 3053-3061) are dumped-and-verified against the 634 cache (revealed value = 4, or 5 when the tracks run inverted), matching objects 19441-19463/19706-19712/19723-19749.
- Trails are session-state: cleared on logout, wrong-object inspects reveal nothing, attacking the wrong spot plays the fail animation without clearing the trail.

Desert/wood kebbit tracking has no usable reference data anywhere I could find, so it's left for a follow-up. Trail coordinates are transcribed from an older-revision reference (polar trails sit on plane 1) — worth an in-game walk-through of both trails.

## Tests
`KebbitTrackingTest`: deterministic full trail catch with varbit progression asserts, wrong-object inspect, missing noose wand, wrong-spot attack fail, level gate.


HarleyGilpin:
Consolidated into #1239.

PR #1245, opened 2026-08-31
Fix evil twin, pheasant and random event graphic bugs

### LINKED ISSUE #1113: Pinball random event npc dialogue anims
tilt & flippa have broken animations

### LINKED ISSUE #1227: Pheasant random event bugs

HarleyGilpin:
Closes #1227. Replaces #1233, whose branch picked up duplicate commits from an upstream rebase.

## Evil Twin

With `world.npcs.randomWalk` on, the npc tick re-applies `Wander` over `EmptyMode` every tick, so a grabbed suspect strolled around mid-lift and wandered back out of the jail while the face-north flag was applied. WorldTest forces `randomWalk` off, which is why no test caught it. The grab, the drop at the jail and the re-entry spawn now use `PauseMode`, and the twin turns north on the same tick as the jailed animation, a tick after `clearTransform`, so the face update can't be lost to the transform or the walk's leftover facing.

New test runs the grab with `randomWalk` enabled and asserts the twin ends on the jail tile facing north.

## Freaky Forester

* Pheasants die to a single hit from any attack style, even when the accuracy roll misses, handled in `Target.damageLimitModifiers` alongside `damage_cap` and `immune_death`.
* They also can't deal damage. `Damage.maximum` reads `max_hit_melee` off the npc definition ahead of the combat definition's range, so the inherited `max_hit_melee = 10` had to go for the combat definition's `target_hit` of 0 to apply.
* The defend anim was wrong because pheasants used the chicken combat definition, and the chicken anims (5387-5389) belong to the reworked chicken rig the pheasant model never got. Pheasants now have their own combat definition with the old bird rig anims (2371-2373) and pheasant sounds (690-692).

New test forces every accuracy roll to fail and asserts one swing kills the pheasant and drops the raw bird; it fails without the fix.

## Random event puff of smoke

Adds a shared `eventHerald` for the spawn and `poof` for the removal, and routes every event through them. `poof` spawns the graphic on the npc's tile rather than flagging it on the npc, the way `Imp` already does, so removal can't swallow it.

New test asserts the graphic is still flagged on the tick the npc reaches clients, since the spawn is queued a tick ahead of that.

## Area exit handlers on the way into an event

Reported after the above: being taken by a random event out of Mort Myre kept the swamp decay running, and out of the barrows crypts kept the barrows overlay on screen for the whole event.

`Movement.move` subtracted the player's `instance_offset` from both ends of a move. The six events that build an instance set that offset before the teleport that carries the player in, so the `from` tile - still real map - was offset into a tile the player was never on, and no area they actually left fired its exit handler. Each end is now mapped on its own, and only when it is inside the block of regions instances are allocated from.

Fight cave was relying on that asymmetry. Its exit tile sits inside `tzhaar_fight_cave_multi_area`, so leaving on foot is not an area change and never was one; the instance end just used to be compared against a meaningless tile. It now tears the instance down in `leave`, and the exit handler keeps the logout path it always had.

Swamp decay had no exit handler at all and checked immunity ahead of the area, so a player carrying a blessed sickle out of the swamp kept the timer, the message and the shield graphic for the rest of their session however they left.

New tests kidnap a player out of the barrows crypts and out of Mort Myre and assert the overlay closes and the timer stops; both fail without the fix.

## Fight cave progress lost on shutdown

`World.shutdown` despawned every player without marking them as leaving the world, so every area exit handler took its walked-out branch. Fight cave cleared `fight_cave_wave` and left the player standing on an instance tile, and the shutdown save persisted both - a restart mid-cave cost the run and logged the player back in inside a dead instance. Godwars killcounts and greegrees were dropped the same way.

`logged_out` is now set before the despawn, which is the flag all of those handlers already check. Logging out normally was never affected; `AccountManager.logout` sets it.

New test takes a player into the cave, shuts the world down and asserts the wave survives and the player is back on real map; it fails without the fix.

## Flippa and Tilt chatheads

Both pinball trolls had no `dialogue` key, so `NPCDialogue` fell through to the default expression names. Those are the reworked human chathead set (`expression_neutral1` = 9807 onwards, skeleton 2165); Flippa and Tilt share chathead model 3790, a 2005 troll head rigged to the pre-rework skeleton 82, so the frames tore it apart mid-dialogue.

They now use the `old` family (554-617, skeleton 82), which is the same set void already gives other pre-rework models - frogs, clay golems, etc.

Closes https://github.com/GregHib/void/issues/1113

## Pinball posts

The next post was drawn from all five, so it could land on the post that had just been tagged - which reads as the event having ignored the tag rather than as a new target. The post that was just lit is now excluded from the draw.

Tagging was also silent. The 634 cache carries this event's audio: 2278 is the pillar tag and 2279-2283 are one tone per pillar.

Tests now tag whichever post is flashing rather than assuming it never moves, plus a new case asserting the target never repeats across eight tags.

## Evil Bob fishing spot pan

Sped up the camera pane and fixed a camera lock bug caused by moving after the last tick of the camera pane cutscene.

## Quiz Master seating

`startEvent` faced the player at the Quiz Master's tile on the same tick `kidnap` teleported them in, and a turn flagged on a tick the player moved is swallowed by the movement update - the same thing that was making the evil twin face the wrong way. The seat could end up pointing whichever way the player arrived; it only looked right most of the time because the first intro line re-faces the player at the npc a tick or two later.

The sit now happens a tick after the teleport and faces `Direction.NORTH` explicitly - the podium is always four tiles north of the seat, so facing north is the invariant rather than "wherever the npc lookup happened to land".

The studio is also now instanced, like the maze, the gravedigger's graveyard and Evil Bob's island. It was a shared static region, so two players taking the event at once stood on the same seat tile. Region 7754 is copied per contestant, the Quiz Master is spawned into the copy owned by them, and logging out inside it drops them back where they were taken from. The static Quiz Master spawn is removed with it: instances don't copy npcs, so the studio's npc has to be spawned per instance, and the real region is no longer reachable.

New test walks the event tick by tick and asserts the sit animation lands strictly after the teleport, which is the condition the client needs for the turn to survive. Asserting the final facing alone is not enough: the intro dialogue re-faces the player anyway, so that assertion passes even with the bug in place.

### Clicking out of the quiz

Closing the chat box cancels the question's `IntEntry` suspension (`closeDialogue` nulls it), which orphaned the coroutine running the show - the interface stayed shut, the player stayed sat in the studio, and nothing brought it back, so they were stuck until a relog. Any click that reaches `closeInterfaces()` does it: walking, an object, an npc.

`npcOperate("Talk-to", "quiz_master")` was meant to be the way back in but never worked at all: the contestant sits four tiles from the podium with the desk between them, and an operate interaction needs a tile next to its target, so it aborted before reaching the handler. It is an approach interaction now with the range set from the seat, the same shape as the barmaid behind her bar, which also covers clicking off the intro lines before the first question is up. The npc's `interact_distance = 4` went with it - nothing in the engine reads that key.

An `interfaceClosed` hook now re-shows the question whenever it is closed with an answer still pending, so there is nothing to click out of. Rolling a question is split from showing one, so a stray click puts the *same* question back rather than handing out a fresh roll.

Clicking away also stands the contestant up and turns them, so a resumed show had them answering from their feet. Facing the podium and the chair animation are one step now, applied on the way in and again on every resume.

### Clicking out of the mime performance

The Mime is the only other event that suspends on player input like this, and closing its emote picker cancelled the round's suspension the same way. Walking off was already covered by the event's `walkTrigger`, but any other click - an object, an npc, an interface - was not. The close now restarts the round through the same path the walk trigger uses, keeping the score; both can fire on the same tick, so the queued restart is replaced rather than stacked.


game/src/main/kotlin/content/activity/event/random/quiz_master/QuizMaster.kt
```diff
@@ -60,8 +62,12 @@ class QuizMaster : Script {
         }
         quizHerald()
         kidnap(ROOM)
+        // A turn flagged on the teleport tick is lost to the movement update, leaving the player
+        // sat facing whichever way they arrived. Turn on the tick the sit animation plays instead;
+        // the studio isn't instanced, so the seat is always the same one and the podium is north.
```
GregHib: A lot of these comments are quite excessive and could be removed

game/src/main/kotlin/content/area/karamja/tzhaar_city/TzhaarFightCave.kt
```diff
@@ -101,16 +101,20 @@ class TzhaarFightCave(
...
-                val offset = tile.delta(tile.region.tile)
-                tele(region.tile.add(offset))
-            } else {
+            if (!get("logged_out", false) || wave == -1) {
                 clear("fight_cave_wave")
+                return@exited
             }
+            // Save the player's relative position in the original region
+            val offset = tile.delta(tile.region.tile)
+            tele(region.tile.add(offset))
```
GregHib: Nothing changed?

game/src/main/kotlin/content/activity/event/random/quiz_master/QuizMaster.kt
HarleyGilpin: I stripped out the comments. https://github.com/GregHib/void/pull/1245/commits/ad4213b8b9c94ed3a22d230dfd151135efd6d2fc

game/src/main/kotlin/content/area/karamja/tzhaar_city/TzhaarFightCave.kt
HarleyGilpin: Reverted in 393b15e14b1f102b3fa51ffb112ab52b047b7717.

GregHib (approved):

PR #1256, opened 2026-09-02
Add the Evil Tree Distraction & Diversion

HarleyGilpin:
Builds on the existing `evil-tree` work in progress and finishes it so the event can be played end to end: a seedling sprouts somewhere in the world, grows into a tree, gets chopped and burned down, and pays out rewards.

image

Reference: [2011 wiki](https://runescape.wiki/w/Evil_Tree?oldid=4518377) for the numbers and the flow.

### Event lifecycle

Two world timers drive the event — `evil_tree_spawn` schedules the next tree, `evil_tree` ticks the live one. Timers rather than `World.queue` because clearing a queue invokes its pending block, which would respawn a tree during teardown.

A tree sprouts two minutes after startup and two hours after the last one died, both configurable, and its location is announced when `world.messages` is on.

### Growth

Saplings advance a stage every `seed_health` growth points, gained either from the growth timer or from nurturing, which pays the tier's farming experience spread across the stage. Roots briefly shoot out of the ground near a sprouting sapling, once per player per tree. The final stage displaces anyone standing where the 3x3 tree appears, sets the tree's health and bursts four roots out of the ground.

### Chopping and burning

* Chopping the trunk uses the woodcutting success rates of the matching log and takes a point of health per successful chop; the tree swaps to its damaged and dying models at two thirds and one third health.
* Chopping a root yields evil tree kindling and dies after three to seven successful chops. A replacement root bursts out every minute, shoving and stunning anyone stood on top of it.
* Lighting kindling against the trunk places a fire in one of the twelve spots around the tree; every lit fire burns a point of health per tick cycle.
* Lightning strikes at ten, twenty and thirty minutes, capping the tree at half then a quarter of its health before felling it outright.

### Rewards

Kindling is the currency. It's handed in to the leprechaun or taken from the reward stump (varbit 1542 picks the reward or blank stump per player), capped at two hundred a day, paying coins, noted logs and "evil tree magic" scaled by the amount handed in. Evil tree magic halves the chance of felling a tree, escorts logs to the bank once the inventory is full and doubles the bird's nest rate.

Players can help with two trees a day; inspecting a tree doesn't count towards that.

### Fixes to the work in progress

* The growth timer registered player handlers (`timerStart`/`timerTick`) against an npc timer, so the body never ran and a zero interval task span every tick.
* A string concatenation in the stage `when` (`"evil_tree_young" + "evil_tree_young_large"`) meant neither young stage matched.
* The burning code checked and removed `kindling`, the Lunar Diplomacy item, rather than `evil_tree_kindling`.
* `chop` and `light` were never wired to an option, so a single click advanced a whole stage with no level check, animation or experience.
* `wander_radius` isn't a valid npc parameter and threw on startup; wander ranges live in the `npc_ranges` table.
* `tree_xp` held the wiki's root experience values; corrected both columns against the wiki, along with elder's burn experience and woodcutting level.
* The spirit tree referenced an object id (`strange_sapling`) that doesn't exist, so its sapling branch was unreachable, and its countdown was a placeholder. Both now work off the live event state.

### Testing

`EvilTreeTest` covers growth by nurture and by timer, the growth spurt, chopping the trunk and the roots, root death, burning kindling, fire damage, the lightning strikes, death and respawn scheduling, the reward stump, the tool loan, and both daily limits. Full suite passes.


game/src/main/kotlin/content/activity/evil_tree/EvilTreeState.kt
```diff
@@ -0,0 +1,86 @@
...
+    var grownTick: Long = NOT_GROWN
+    var strikes: Int = 0
+    var dead: Boolean = false
+    var spawnId: Int = 0
+    var growthDelay: Int = 0
+    var rootDelay: Int = 0
+    var deathDelay: Int = 0
+    var respawnTicks: Int = 0
+    val roots = mutableMapOf<String, Root>()
+    val fires = mutableMapOf<String, GameObject>()
```
GregHib: As 80% of this state isn't used outside of EvilTree.kt can we move all of it into there, and anything that is accessed externally into EvilTree.Companion ?

game/src/main/kotlin/content/activity/evil_tree/EvilTreeLeprechaun.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+    if (this["evil_tree_trees", 0] >= Settings["events.evilTree.dailyTreeLimit", 2]) {
+        message("You've already helped with as many evil trees as you can today.")
+        return false
+    }
+    this["evil_tree_spawn_id"] = EvilTreeState.spawnId
+    inc("evil_tree_trees")
+    this["evil_tree_rewards"] = true
+    return true
+}
```
GregHib: These methods need specific names as they are global. `evilTreeTimerReset` & `hasEvilTreeLimit` perhaps?

game/src/main/kotlin/content/skill/woodcutting/Woodcutting.kt
```diff
@@ -183,3 +190,23 @@ class Woodcutting(val drops: DropTables) : Script {
...
+
+fun calculateChance(hatchet: Item, treeHatchetDifferences: IntRange): Int = (0 until hatchet.def["rank", 0]).sumOf { calculateHatchetChance(it, treeHatchetDifferences) }
+
+/**
+ * Calculates the chance of success out of 256 given a [hatchet] and the hatchet chances for that tree [treeHatchetDifferences]
+ * @param hatchet The index of the hatchet (0..7)
+ * @param treeHatchetDifferences The min and max increase chance between each hatchet
+ * @return chance of success
+ */
+fun calculateHatchetChance(hatchet: Int, treeHatchetDifferences: IntRange): Int = if (hatchet % 4 < 2) treeHatchetDifferences.last else treeHatchetDifferences.first
```
GregHib: Put in companion object please, they don't need global scope

GregHib (changes requested):

game/src/main/kotlin/content/activity/evil_tree/EvilTreeState.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/d002509394dd1f4820909abfd4942e2bbf6306ac - EvilTreeState is gone. What the leprechaun and spirit tree read (place, type, spawnTile, spawnId, active, alive, sapling, centre) is on EvilTree.Companion, the rest are fields on the script instance (internal so the tests can reach them via the script instance).

game/src/main/kotlin/content/activity/evil_tree/EvilTreeLeprechaun.kt
HarleyGilpin: Renamed to evilTreeDailyReset and evilTreeInteract in https://github.com/GregHib/void/commit/d002509394dd1f4820909abfd4942e2bbf6306ac. Went with evilTreeInteract over hasEvilTreeLimit as it returns true when the player is allowed in and also counts the tree towards their limit. Moved them next to evilTreeMagic in EvilTree.kt.

game/src/main/kotlin/content/skill/woodcutting/Woodcutting.kt
HarleyGilpin: Moved into the Woodcutting companion in https://github.com/GregHib/void/commit/d002509394dd1f4820909abfd4942e2bbf6306ac, with success back to its original name.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+    /**
+     * Roots briefly shoot out of the ground around a sprouting sapling, once per player per tree.
+     */
+    private fun notice() {
+        Players.forEachInRadius(centre, NOTICE_RADIUS) { player ->
+            if (player["evil_tree_noticed", 0] == spawnId) {
+                return@forEachInRadius
+            }
+            player["evil_tree_noticed"] = spawnId
+            player.gfx("evil_root")
```
GregHib: Should be an areaGfx()

GregHib:
```suggestion
            areaGfx("evil_root", player.tile)
```

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+        }
+    }
+
+    /**
+     * Moves anyone standing where the tree is about to grow out of the way.
+     */
+    private fun displace() {
+        for (tile in spawnTile.toCuboid(3, 3)) {
+            for (player in Players.at(tile)) {
+                push(player, centre)
```
GregHib:
```suggestion
                push(player, centre)
                player.face(centre)
```

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+        }
+        val row = Tables.get("evil_branches").rows().firstOrNull { it.obj("spawn") == spawn.id } ?: return
+        if (roots.containsKey(row.rowId)) {
+            return
+        }
+        val root = GameObjects.add(spawn.id.removeSuffix("_spawn"), spawn.tile, rotation = spawn.rotation)
+        roots[row.rowId] = Root(root, random.nextInt(ROOT_LIFE.first, ROOT_LIFE.last + 1))
+    }
+
+    private fun burst(tile: Tile) {
```
GregHib, line 506: Bursting fails if the player is lighting a fire

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+
+        timerTick("evil_tree_buff") {
+            if (dec("evil_tree_buff") <= 0) Timer.CANCEL else Timer.CONTINUE
+        }
+
+        timerStop("evil_tree_buff") {
+            message("${Colours.DARK_RED.toTag()}Your evil tree magic has worn off.")
+        }
+
+        adminCommand("eviltree", stringArg("minutes"), desc = "Start a new evil tree event in [minutes]", handler = ::command)
```
GregHib:
```suggestion
        adminCommand("evil_tree", stringArg("minutes"), desc = "Start a new evil tree event in [minutes]", handler = ::command)
```

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+                message("You need a tinderbox in order to light a fire.")
+                break
+            }
+            if (!inventory.contains("evil_tree_kindling")) {
+                message("You don't have any kindling to burn.")
+                break
+            }
+            val spot = freeFire() ?: run {
+                message("There's nowhere left to light a fire.")
+                null
```
GregHib: Should be the same as the tile where the player is standing and if it's already on fire: `"That part of the tree is already on fire!"`

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+            pause(LIGHT_TICKS)
+            if (!alive || deathDelay > 0 || fires.containsKey(spot.rowId)) {
+                continue
+            }
+            if (!inventory.remove("evil_tree_kindling")) {
+                break
+            }
+            exp(Skill.Firemaking, row.int("burn_xp") / 10.0)
+            val tile = spawnTile.add(spot.int("deltaX"), spot.int("deltaY"))
+            fires[spot.rowId] = GameObjects.add("evil_tree_fire", tile, rotation = spot.int("dir"), ticks = row.int("fire_life"))
```
GregHib: Fires don't seem to spawn?

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+        }
+        val root = GameObjects.add(spawn.id.removeSuffix("_spawn"), spawn.tile, rotation = spawn.rotation)
+        roots[row.rowId] = Root(root, random.nextInt(ROOT_LIFE.first, ROOT_LIFE.last + 1))
+    }
+
+    private fun burst(tile: Tile) {
+        Players.forEachInRadius(tile, 1) { player ->
+            player.message("You dive out of the way as a new root bursts from the ground.")
+            push(player, tile)
+            player.anim("step_back_startled")
```
GregHib: TODO: check YT if it's this anim or 6999

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+            player["delay"] = STUN_TICKS
+            player.start("stunned", STUN_TICKS)
```
GregHib:
```suggestion
            player.start("stunned", STUN_TICKS)
            target.gfx("stun_long")
```

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
```diff
@@ -0,0 +1,765 @@
...
+            player.start("movement_delay", STUN_TICKS)
+        }
+    }
+
+    private fun killRoot(row: String) {
+        val root = roots.remove(row) ?: return
+        GameObjects.remove(root.obj)
+    }
+
```
GregHib, line 517: Evil roots should also sweep you away if you're standing next to them when active and chopping or lighting fires on the roots or the tree.

https://youtu.be/T9QN_hwh7g0?si=t=26

GregHib (changes requested):

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Confirmed that it is animation id 6999 - https://youtu.be/42bwAlV3IHo?t=141

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Fixed in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Found it: object 11425 only has models for shapes 0 and 4 (wall / wall decoration) in the cache, and it was being added with the default centre piece shape so the client had nothing to draw. It was also fighting the root for the same ground layer on that tile. Fixed in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1 by adding it as a wall decoration facing the tree.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Switched to 6999 in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1 as evil_root_dodge.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Done in https://github.com/GregHib/void/commit/563383e11f8e634928882c651b6e09e8ab4e19a1. Each swing or light attempt checks whether the player is within 2 tiles of the tree centre and next to a settled root, and throws them back with the same dodge and stun as a bursting root. With all four roots up only the corner tiles are safe, so a side has to be cleared before the tree can be chopped or lit from it.

HarleyGilpin: Follow-up in https://github.com/GregHib/void/commit/21a2f8d1632a304e1044548cdca6604fe39a28a7: chopping a root itself never sweeps you, only working on the tree beside one does.

game/src/main/kotlin/content/activity/evil_tree/EvilTree.kt
HarleyGilpin: Follow-up in https://github.com/GregHib/void/commit/1cb88f919dd9544bb4acb8a56ad6b06c3cdfc181: the flame wasn't animating either. Animation 9593 lives on the varbit multilocs (14887/15253...) rather than on 11425, so it's now played on the object when lit and replayed to players entering the zone afterwards.

HarleyGilpin: Second attempt in https://github.com/GregHib/void/commit/fc6d620d8e77b4ec0805af9926753d8c34fb4390. The client only creates an animated scene object when the def has an animation, transforms or a varbit, and the animate-object packet is ignored for anything else, so a plain 11425 can never move. The map uses eight varbit multilocs around the tree (14887/15253 and siblings, one per side and corner) that carry anim 9593, so the fires are now those objects with per-player varbits kept in sync on light, burn out, tree death and zone entry. The fire is picked by which side or corner of the tree the player is standing on.

HarleyGilpin: Fixed, fires can now bit lit around the tree.

GregHib (approved):

PR #1263, opened 2026-09-04
add makefile

MaatheusGois:
Adds a top-level `Makefile` so contributors can start/stop/status the server without remembering the gradle wrapper invocation.

Currently the documented path is `./gradlew :game:run` directly, with no way to background it cleanly or check whether it's already running. This wraps the lifecycle in five recipes:

- `make start` — start the server in the background, wait for port 43594 to listen, tail the log
- `make stop` — gracefully shut down (SIGTERM → 15s grace → SIGKILL), clearing any leftover listener
- `make status` — show pid + port state + last log lines
- `make restart` — `stop` + `start`
- `make logs` — `tail -f server.log`
- `make clean` — wipe `*/build` and `.gradle` (fixes the "Storage for [.../class-attributes.tab] is already registered" failure when a previous build died mid-write)

`START_TIMEOUT` overrides the readiness wait (default 600s) — `make start START_TIMEOUT=900`.

## Notable behavior

- `start` uses `setsid nohup ...</dev/null` so the gradle worker's child JVM is in a new session and survives the parent shell exit (the previous `nohup ... &` form kept getting SIGHUP'd when `make` returned).
- `stop` and the port-cleanup fallback both verify the saved pid is still a `java`/`bash`/`sh` process before signalling, so a recycled pid or an unrelated listener won't get killed.

## Draft because

I want a maintainer sanity check on whether a Makefile belongs in the repo at all (some projects prefer a script in `scripts/` or a launchd plist), and on the chosen lifecycle approach before this gets merged. Happy to restructure — e.g. drop `start`/`stop`/`status` and only keep `clean`, since `./gradlew :game:run` is one line — if that's preferred. The actual gradle commands are unchanged from what the README already recommends.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,143 @@
+# Void - RuneScape 634 server
+# Game protocol port (NOT http) — access with void-client on localhost:43594
+
+PORT      := 43594
+PID_FILE  := .server.pid
+LOG_FILE  := server.log
+
+# Override on the command line, e.g. `make start START_TIMEOUT=900`.
+# Default 10 min covers a clean :game:compileKotlin from cold (~4 min build
+# + engine startup) without silencing real failures.
+START_TIMEOUT ?= 600
+
+.PHONY: start stop restart status logs help clean
+
+help:
+	@echo "Void server commands:"
+	@echo "  make start    - build and run server in background"
+	@echo "  make stop     - stop the server"
+	@echo "  make status   - show if server is running"
+	@echo "  make restart  - stop and start again"
+	@echo "  make logs     - follow server output (Ctrl+C to exit)"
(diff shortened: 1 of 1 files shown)


GregHib:
> I want a maintainer sanity check on whether a Makefile belongs in the repo at all

Yeah that's what I was thinking, makefile's are more C world to me it's not something I come across very often in JVM, gradle you can use of course but it's better to use it built into your IDE; though I can appreciate not everyone wants to use IntelliJ like me.

Though the start/stop commands are nice for developers I kinda of want to discourage those who don't know what they're doing from installing programs and executing arbitrary commands that they won't know the consequences of, and for easy setups point them to releases instead. Though you could argue ofc that gradle is just abstracting that away, but I am more familiar and comfortable with it at least.

PR #1264, opened 2026-09-05
Add missing potion drink effects

HarleyGilpin:
Fills in the drink interactions for the potions that had item definitions but no effect, and fixes the bugs found while auditing them.

Six commits, each self-contained and green on its own.

## Bugs

Several potions were not drinkable at all. `strength_potion_2` had neither `heals` nor `excess`, so it was rejected by the guard at the top of `Eating.consume`. `antidote+_mix_2` decanted into `super_antipoison_mix_1`, turning the item into a different potion when drunk. The `extreme_magic` block had its `weight` and `examine` values reversed across all four doses, so the 4-dose potion described itself as one dose.

`weak_range_potion` was registered for an item actually named `weak_ranged_potion`. Wildcard patterns that match nothing throw at startup, but literal ids that match nothing register silently, which is why this went unnoticed.

Energy potions assigned a tenth of the player's *current* run energy instead of adding a tenth of the maximum, so drinking one drained you. Nettle tea had the same shape.

## Effect matching

`potionEffects` matched with `startsWith`, which made every effect depend on branch order. `antipoison` was tested before `antipoison+` and `antipoison++`, so antidote+ and antidote++ granted 90 seconds of immunity instead of 9 and 12 minutes. The same flaw silently dropped any potion whose name did not begin with its family's prefix, including `super_prayer` and `antidote+_mix`.

Matching now happens on the item id with its dose suffix removed, so families are selected by exact name and branch order means nothing.

The dose handler treated any consumable ending `_1` to `_4` as a potion, so cooked crab meat and stealing creation food reported doses remaining. It now recognises potions by their empty vial, which also covers 5-dose potions that the old range excluded.

## New effects

Extreme potions and super prayer. Overload grants the boost of all five extremes at once, so the two share one definition rather than repeating the amounts where they could drift apart.

Stealing creation and castle wars potions were undrinkable, missing both a dose chain and an effect. Each reuses the effect of the potion it copies. The examine text identifies the tier: the stealing creation ranging, magic, defence, prayer and energy potions are ordinary strength, while its attack and strength potions and all six castle wars potions are supers.

The poison chalice was consumed and vanished without doing anything. It now leaves a cocktail glass and applies one of seven random outcomes, from a slight taste of apricot through stat drains and heals to a large boost, or a small boost followed by damage. Both healing outcomes scale with maximum life points.

All ten Herblore Habitat potions now work. The mining and woodcutting potions do not bank every resource; each ore or log gathered has a chance to open a short window during which they go to the bank instead of the inventory. The farming potion gives a chance of an extra herb on herb patches only, and the fishing potion turns some raw sharks into raw baron sharks.

## Testing

`PotionEffectsTest`, `PoisonChaliceTest` and `JujuTest` are new, covering every bug above as a regression plus the new effects. One asserts that overload and an extreme ranging potion produce the same boost, so the shared definition cannot drift. Full suite passes: 411 classes, 3513 tests.


data/minigame/castle_wars/castle_wars.items.toml
```diff
@@ -358,121 +358,169 @@ kept = "Reclaim"
 [cw_super_attack_potion_4]
 id = 18715
 weight = 0.035
+excess = "cw_super_attack_potion_3"
+empty = "vial"
```
GregHib: Not sure if these would become vials vs just breaking?

data/minigame/stealing_creation/stealing_creation.items.toml
```diff
@@ -641,6 +641,8 @@ examine = "The sacred clay has transformed into a wood-like substance."
 [prayer_potion_stealing_creation_5]
 id = 14207
 weight = 0.5
+excess = "prayer_potion_stealing_creation_4"
+empty = "vial"
```
GregHib: Same with these, do they break or become a special stealing creation empty vial?

HarleyGilpin:
Two notes for whoever reviews this.

**The juju rates are not verified for this revision.** The 11% banking proc, the 75 tick window, the 500 tick effect duration and the 30% shark conversion all come from a later revision's implementation, and I could not find period documentation for them. They are named constants at the top of `Juju.kt` so they can be corrected in one place without touching the hooks. The shape of the effect (a proc that opens a window during which gathered resources bank themselves) is the behaviour from the potion's release until it was reworked, so that part should be right for 634 even if the numbers need adjusting.

**Two mappings look wrong but are not.** `magic_potion_stealing_creation` uses the `super_magic_potion` boost and `ranging_potion_stealing_creation` uses `super_ranging_potion`. That reads oddly because items 3040 and 2444 are named `super_magic_potion` and `super_ranging_potion` here, while the cache calls them plain "Magic potion" and "Ranging potion". The stealing creation copies are the ordinary tier and map to the same effect as the items they copy. Renaming those two families to match the cache would be worth doing, but it touches item configs, recipes and shop data, so I left it out of this branch.

data/minigame/castle_wars/castle_wars.items.toml
HarleyGilpin: You're right, they break. Fixed in f41926cde.

data/minigame/stealing_creation/stealing_creation.items.toml
HarleyGilpin: Same, they break. Also fixed in f41926cde.

game/src/main/kotlin/content/skill/constitution/drink/Juju.kt
```diff
@@ -0,0 +1,100 @@
...
+ */
+fun Player.jujuBank(effect: String, item: String, amount: Int): Boolean {
+    if (jujuActive(effect) && random.nextInt(100) < BANK_PERCENT) {
+        set("${effect}_bank", BANK_DURATION)
+        timers.restart("${effect}_bank")
+    }
+    if (!jujuActive("${effect}_bank")) {
+        return false
+    }
+    if (!bank.add(item, amount)) {
```
GregHib, line 57: Can we use BankDeposit.deposit() for safety

game/src/main/kotlin/content/skill/constitution/drink/PotionEffects.kt
```diff
@@ -20,85 +23,139 @@ import java.util.concurrent.TimeUnit
...
+            if (poisoned) {
+                poisonDamage -= 10
+            }
+            runEnergy += (MAX_RUN_ENERGY / 100) * 5
+            levels.boost(Skill.Constitution, 50, maximum = 50)
+        }
+        // Registered so a dose still reports itself, but their effects are not implemented.
+        effect("juju_cooking_potion", "juju_hunter_potion") { }
+        // Used on something else rather than drunk, so drinking them does nothing.
+        effect("compost_potion", "guthix_balance", "serum_207", "serum_208", "olive_oil", "sacred_oil") { }
```
GregHib: Can be removed as not in use as there's no drink option?
```suggestion
```

game/src/main/kotlin/content/skill/constitution/drink/PotionEffects.kt
```diff
@@ -20,85 +23,139 @@ import java.util.concurrent.TimeUnit
...
+        // Drinkable, but with no effect of their own.
+        effect("guthix_rest") {
+            if (poisoned) {
+                poisonDamage -= 10
+            }
+            runEnergy += (MAX_RUN_ENERGY / 100) * 5
+            levels.boost(Skill.Constitution, 50, maximum = 50)
+        }
+        // Registered so a dose still reports itself, but their effects are not implemented.
+        effect("juju_cooking_potion", "juju_hunter_potion") { }
```
GregHib, line 150: TODO?

game/src/main/kotlin/content/skill/constitution/drink/PotionEffects.kt
```diff
@@ -20,85 +23,139 @@ import java.util.concurrent.TimeUnit
...
+
+/**
+ * Every dosed potion, keyed by its item id with the dose suffix removed. Membership is what marks
+ * an item as a potion, so a potion with no effect yet is still registered with an empty one.
+ */
+object PotionEffects {
+    val effects: Map<String, Player.() -> Unit> = buildMap {
+        effect("antifire", "antifire_mix") { antifire(6) }
+        effect("super_antifire") { superAntifire(6) }
+        effect("overload") {
```
GregHib, line 60: Rather than having a map with this custom mutation method, it could be just
```
fun Player.potionEffects(potion: String): Boolean {
  when (potion.substringBeforeLast('_')) {
    "antifire", "antifire_mix" -> antifire(6)
    // ...
    else -> return false
  }
  return true
}
```

game/src/main/kotlin/content/skill/constitution/drink/Potions.kt
```diff
@@ -3,26 +3,26 @@ package content.skill.constitution.drink
...
+        consumed("*") { item, _ ->
+            val separator = item.id.lastIndexOf('_')
+            if (separator == -1) {
                 return@consumed
             }
-            val doses = item.id.last().digitToInt()
-            if (doses != 1) {
-                message("You have ${doses - 1} ${"dose".plural(doses - 1)} of the potion left.")
-                potionEffects(item.id)
+            val doses = item.id.substring(separator + 1).toIntOrNull()
```
GregHib, line 15:
```suggestion
            val doses = item.id.substringAfterLast('_').toIntOrNull()
```

game/src/main/kotlin/content/skill/constitution/drink/Potions.kt
```diff
@@ -3,26 +3,26 @@ package content.skill.constitution.drink
...
+            if (doses == null || doses !in 1..5) {
+                return@consumed
+            }
+            if (!PotionEffects.effects.containsKey(item.id.substring(0, separator))) {
                 return@consumed
             }
-            message("You have finished your potion.")
-            if (contains("smash_vials")) {
-                inventory.remove(slot, item.id)
-                message("You quickly smash the empty vial using the tick a Barbarian taught you.")
```
GregHib, line 25: Can we put vial smashing back?

GregHib (changes requested):

game/src/main/kotlin/content/skill/constitution/drink/Juju.kt
HarleyGilpin: Done in 8c19b55af. The ore/log now goes into the inventory and out again through `BankDeposit.deposit(check = false)`, so it gets the unbankable and full-bank handling. If nothing left the inventory the resource is taken back out and the window closes as before.

game/src/main/kotlin/content/skill/constitution/drink/PotionEffects.kt
HarleyGilpin: Added a TODO in 5c369a8ea.

game/src/main/kotlin/content/skill/constitution/drink/PotionEffects.kt
HarleyGilpin: Done in 5c369a8ea. `potionEffects` is now the `when` and returns whether the id matched, which is what `Potions` was using the map keys for. The registry test now drinks each dosed base name and checks for a false return instead of reading the key set.

game/src/main/kotlin/content/skill/constitution/drink/Potions.kt
HarleyGilpin: Applied in 5c369a8ea.

game/src/main/kotlin/content/skill/constitution/drink/Potions.kt
HarleyGilpin: That block was dead before this branch touched it. It read `smash_vials`, but the Barbarian Outpost dialogue sets `vial_smashing` (BarbarianGuard.kt:77), and Eating.kt already handles that flag: with `vial_smashing` set and `excess = "vial"` it removes the item and prints the smash message. So smashing still works on this branch through Eating.kt, and re-adding it here would run the removal twice. The unused `[smash_vials]` entry in barbarian_outpost.vars.toml can go too if you want, happy to drop it.

HarleyGilpin: Turned out the definition was the broken half, not the code: `vial_smashing` had no entry at all, so the flag the guard sets never persisted. Renamed it in 3cdcf02c7 and added tests for the smash and the persist.

GregHib (approved):

PR #1266, opened 2026-09-06
Fix Zaryte bow weapon style

MaatheusGois:
The Zaryte bow was missing weapon_style = 16. This item-definition-only fix classifies it as ranged instead of melee.


GregHib:
image
Closing as it looks like it already has `weapon_style = 16` baked into the cache? If this was solve a problem please raise an issue separately

PR #1274, opened 2026-09-07
Add the Duel Arena minigame

HarleyGilpin:
Adds the Duel Arena - `content/minigame/duel_arena`

## What's included
- **Challenge** player option on the walkways/hospital (interface 640 friendly/staked selection, chat request, mutual-accept handshake like trading).
- Rules screens 631 (staked) / 637 (friendly): all 12 rules plus the 11 equipment bans, mirrored to both players through the real 634 varbits, with the No Ranged/Melee/Magic and Obstacles/No Movement exclusions.
- Staking through the 628 side tab into inventory 134, mirrored to the opponent's secondary container; tradeable items only.
- Confirmation screens 626 / 639 with the before/during lists and stake summaries, then teleport into a random obstacle or obstacle-free arena (side-by-side placement for No Movement), banned gear unequipped, countdown "3, 2, 1, FIGHT!".
- Enforcement: duellers and their familiars may only attack each other and only after the countdown; combat style, fun weapon, special attack, food, drink, prayer, summoning, movement, equipment, teleport, drop and logout blocks.
- Ending by death, the forfeit trapdoor or disconnect: both players healed and sent to the hospital, winner gets 633/634 with the winnings in inventory 136 (claimed on close, bank fallback), audit log entries, and an in-memory scoreboard of the last 50 duels on object 3192.

Generic hooks were added to `Eating`, `Equipping`, `Movement`, `Summoning` and `QuickPrayers` (`no_food_message`, `blocked_equip_slots`, ...) because item/interface option handlers can't veto each other.

## Tests
`DuelTest` (21 WorldTest cases) covers the request handshake, rule toggles and exclusions, stake mirroring/removal/restrictions, accept space and fun-weapon checks, countdown and fight start, death and forfeit payouts, disconnect forfeit, logout veto, walk-away decline, third-party attack blocking, leftover-item recovery and the scoreboard cap. Full `:game:test` passes.

### DIFF
--- a/game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
+++ b/game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
@@ -45,7 +45,7 @@ class DuelArena : Script {
     }
 
     companion object {
-        const val CHALLENGE_SLOT = 3
+        const val CHALLENGE_SLOT = 1
 
         /**
          * Moves everything in [inventory] to the player's inventory, falling back to their bank
--- a/game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
+++ b/game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
@@ -1,11 +1,14 @@
 package content.minigame.duel_arena
 
-import content.entity.player.bank.bank
+import com.github.michaelbull.logging.InlineLogger
+import content.entity.player.bank.BankDeposit
 import world.gregs.voidps.engine.Script
 import world.gregs.voidps.engine.client.message
 import world.gregs.voidps.engine.client.ui.close
 import world.gregs.voidps.engine.client.ui.open
+import world.gregs.voidps.engine.data.SaveQueue
 import world.gregs.voidps.engine.entity.character.player.Player
+import world.gregs.voidps.engine.get
 import world.gregs.voidps.engine.inv.Inventory
 import world.gregs.voidps.engine.inv.clear
 import world.gregs.voidps.engine.inv.inventory
@@ -41,11 +44,13 @@ class DuelArena : Script {
             if (inventories.contains("duelwinnings")) {
                 returnItems(winnings)
             }
+            save(this)
         }
     }
 
     companion object {
         const val CHALLENGE_SLOT = 1
+        private val logger = InlineLogger()
 
         /**
          * Moves everything in [inventory] to the player's inventory, falling back to their bank
@@ -53,18 +58,32 @@ class DuelArena : Script {
         fun Player.returnItems(inventory: Inventory) {
             var banked = false
             for (index in inventory.indices) {
-                if (inventory[index].isEmpty()) {
+                val item = inventory[index]
+                if (item.isEmpty()) {
                     continue
                 }
                 if (inventory.move(index, this.inventory)) {
                     continue
                 }
-                if (inventory.move(index, bank)) {
+                BankDeposit.deposit(this, inventory, item, item.amount, index, check = false)
+                if (inventory[index].isEmpty()) {
                     banked = true
+                } else {
+                    logger.warn { "Unable to return duel item $item to $this" }
                 }
             }
             if (banked) {
                 message("Some of your duel items have been sent to your bank.")
+            }
+        }
+
+        /**
+         * Persist both sides straight after items change hands so a crash can't restore a stale stake
+         */
+        fun save(vararg players: Player) {
+            val queue: SaveQueue = get()
+            for (player in players) {
+                queue.save(player)
             }
         }
     }
--- a/game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
+++ b/game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
@@ -1,0 +1,71 @@
+package content.minigame.duel_arena
+
+import content.entity.player.bank.bank
+import world.gregs.voidps.engine.Script
+import world.gregs.voidps.engine.client.message
+import world.gregs.voidps.engine.client.ui.close
+import world.gregs.voidps.engine.client.ui.open
+import world.gregs.voidps.engine.entity.character.player.Player
+import world.gregs.voidps.engine.inv.Inventory
+import world.gregs.voidps.engine.inv.clear
+import world.gregs.voidps.engine.inv.inventory
+import world.gregs.voidps.engine.inv.move
+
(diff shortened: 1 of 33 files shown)


game/src/main/kotlin/content/minigame/duel_arena/DuelRulesScreen.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        /**
+         * Both players are checked so either side sees why the duel can't start
+         */
+        fun canAccept(duel: Duel): Boolean {
+            for (player in duel.players) {
+                val opponent = duel.opponent(player)
+                if (duel.hasRule("fun_weapons") && !DuelRules.hasFunWeapon(player)) {
+                    player.message("Fun Weapons is selected but you don't have a 'fun weapon'.")
+                    opponent.message("Fun Weapons is selected but your opponent does not have a 'fun weapon'.")
```
GregHib: This should go to the player not the opponent

game/src/main/kotlin/content/minigame/duel_arena/DuelRulesScreen.kt
```diff
@@ -0,0 +1,142 @@
...
+                    return false
+                }
+                var needed = DuelRules.removedEquipment(player, duel).size
+                if (duel.staked) {
+                    needed += player.stake.count + opponent.stake.count
+                }
+                if (needed > player.inventory.spaces) {
+                    player.message("You do not have enough space for the items removed and/or the stake.")
+                    opponent.message("Your opponent does not have enough space for the items removed and/or the stake.")
```
GregHib:
```suggestion
                    player.message("Your opponent does not have enough space for the items removed and/or the stake.")
```

game/src/main/kotlin/content/minigame/duel_arena/DuelConfirm.kt
```diff
@@ -0,0 +1,115 @@
...
+            val during = during(duel)
+            for (index in 0 until BEFORE_LINES) {
+                player.interfaces.sendText(screen, "before_$index", before.getOrElse(index) { "" })
+            }
+            for (index in 0 until DURING_LINES) {
+                player.interfaces.sendText(screen, "during_$index", during.getOrElse(index) { "" })
+            }
+            if (duel.staked) {
+                player.interfaces.sendText(screen, "stake", itemsText(player.stake))
+                player.interfaces.sendText(screen, "other_stake", itemsText(player.otherStake))
```
GregHib: This causes a ui bug because text doesn't need to be sent?

Image

game/src/main/kotlin/content/minigame/duel_arena/DuelEnd.kt
```diff
@@ -0,0 +1,210 @@
...
+            player.clear("no_drinks_message")
+            player.clear("no_prayer_message")
+            player.clear("no_summoning_message")
+            player.clear("no_movement_message")
+            player.clear("blocked_equip_slots")
+            player.clear("blocked_equip_message")
+            for (rule in DuelRules.all) {
+                player.clear("duel_$rule")
+            }
+            player.attackers.clear()
```
GregHib, line 157: Hits need clearing as well.
If you shoot a killing blow with a bow over distance (aka against a low hp opponent), then quickly forfeit. You lose but the player get's hit after being teleported back to the arena, so they will end up losing their items and being sent to lumbridge.

GregHib (changes requested):

game/src/main/kotlin/content/minigame/duel_arena/DuelRulesScreen.kt
HarleyGilpin: Done in https://github.com/GregHib/void/pull/1274/commits/b2232c0b0. `canAccept` now takes the accepting player and only messages them: "you don't have" for their own missing weapon, "your opponent does not have" for the other side. The opponent hears nothing until they click Accept themselves.

game/src/main/kotlin/content/minigame/duel_arena/DuelRulesScreen.kt
HarleyGilpin: Done in https://github.com/GregHib/void/pull/1274/commits/b2232c0b0, same shape as the fun weapon check: both messages go to the player who clicked Accept. Test updated so the opponent's copy only appears once they press Accept too.

game/src/main/kotlin/content/minigame/duel_arena/DuelConfirm.kt
HarleyGilpin: Right, cs2 206 fills both stake texts on 626 from the offer inventories itself. Removed both sendText calls and the now-unused `itemsText` in https://github.com/GregHib/void/pull/1274/commits/f625a0fc0.

game/src/main/kotlin/content/minigame/duel_arena/DuelEnd.kt
HarleyGilpin: Good catch. `cleanup` now clears each player's `hit` queue in https://github.com/GregHib/void/pull/1274/commits/8f7b83476, so an arrow released before the forfeit never lands after the teleport. Added a test that queues a lethal ranged hit, forfeits through the trapdoor and checks the winner is alive at full HP in the hospital; it fails without the clear.

game/src/main/kotlin/content/minigame/duel_arena/DuelArena.kt
```diff
@@ -0,0 +1,90 @@
...
+            if (inventories.contains("_dueloffer")) {
+                otherStake.clear()
+            }
+            if (inventories.contains("dueloffer")) {
+                returnItems(stake)
+            }
+            if (inventories.contains("duelwinnings")) {
+                returnItems(winnings)
+            }
+            save(this)
```
GregHib, line 47: This is resaving every player every-time they log in, rather than only when things are returned?

game/src/main/kotlin/content/minigame/duel_arena/DuelFight.kt
```diff
@@ -0,0 +1,243 @@
...
+
+        playerLogout {
+            if (duel?.active == true) {
+                message("You can't log out during a duel.")
+                return@playerLogout false
+            }
+            true
+        }
+
+        objectOperate("Forfeit", "duel_arena_forfeit_trapdoor", arrive = false) {
```
GregHib, line 89: How does one forfeit when movement is disabled?

PR #1276, opened 2026-09-07
Fix Ancient Cavern whirlpool delay

MaatheusGois:
Before
Screenshot 2026-09-07 at 14 10 00
After
Screenshot 2026-09-07 at 14 29 57


GregHib (approved):
Works well, thanks

PR #1279, opened 2026-09-07
Fix equipment degradation and broken gear

MaatheusGois:
Fix equipment degradation interval, durability display, depletion alerts, and broken gear restrictions. Tests: ./gradlew :game:test.


MaatheusGois:
Closed in favor of the clean PR based directly on greg/main.

PR #1280, opened 2026-09-07
Fix equipment degradation and broken gear

MaatheusGois:
## Summary

- Slow automatic equipment degradation to the intended interval.
- Show durability and remaining charges to players.
- Notify players when degradable items are depleted.
- Prevent broken weapons and armour from being equipped or used.

## Verification
- ./gradlew :game:test
- Targeted equipment, combat, and degradation tests


GregHib:
Closing because:
1. Broken barrows equipment doesn't have the Wear options
2. Broken nex equipment does have the wear options because they have no stats are are functionally useless
3. Check-charges is already implemented and workin (see DaemonheimRewards.kt)
4. Not sure this actually fixes the timing, just hardcodes it to 100
4. Iirc the correct functionality of a broken item which is still equipped is to just punch

PR #1287, opened 2026-09-10
Creates an option in game.properties to disable player vs player activities

PrettyPrincessPotato:
(Default is enabled)

I saw this wasn't a setting, thought it'd be useful to others too who don't enjoy pvp content. It also just adds another layer of customization for hosts.


GregHib (commented):
Yeah that's a neat idea, thanks! (Will need to be careful with this breaking PvP mini-games in the future but that's not an issue rn as there is none :D )

PR #1290, opened 2026-09-11
Add object examines for a lot of the "most placed" objects

trunksbomb:
Add manually cross-referenced examine texts for a lot of the "most placed" objects (those with 75 or more placements). The reference is a Rune Server thread which is linked in `all_scraped.objs.toml`, where a user scraped examines from live RS2 servers in 2021. This commit adds examine texts for 135 unique objects, representing a total of 40,360 object placements in-world, as well as updating a few other commonly-placed objects that already had definitions in a `objs.toml` file but no examine text.

Honestly it's mostly trees and wheat, but hey now they have descriptions.

### DIFF
--- a/data/entity/obj/all.objs.toml
+++ b/data/entity/obj/all.objs.toml
@@ -675,14 +675,17 @@ examine = "A stand for hats."
 [wheat_zanaris]
 clone = "wheat"
 id = 15506
+examine = "Baby bread."
 
 [wheat_3]
 clone = "wheat"
 id = 15507
+examine = "Baby bread."
 
 [wheat_4]
 clone = "wheat"
 id = 15508
+examine = "Baby bread."
 
 [railing_7]
 id = 15601
--- a/data/area/morytania/port_phasmatys/ectofuntus.objs.toml
+++ b/data/area/morytania/port_phasmatys/ectofuntus.objs.toml
@@ -25,3 +25,4 @@ id = 17118
 [pool_of_slime_4]
 clone = "pool_of_slime"
 id = 17119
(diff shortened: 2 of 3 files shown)


skalfate:
This is a wonderful addition. I am pleased. 😃 The rest is up to Greg to see if the implementation is how its should be implemented.

data/entity/obj/all.objs.toml
```diff
@@ -675,14 +675,17 @@ examine = "A stand for hats."
 [wheat_zanaris]
 clone = "wheat"
 id = 15506
+examine = "Baby bread."
```
GregHib, line 678: These aren't necessary, the examines will come from the cloned object

GregHib (changes requested):
Thanks, we don't usually accept large amounts of definitions (we could ofc dump them all but then it becomes a mountain of unverified data and a huge manual undertaking to validate, so we favour adding them alongside content as we go), however what you've done with the most commonly used, and it's not necessarily objects which fit alongside content, it makes a lot of sense and I'm happy to merge this in once it's ready.

To get it ready it'd need some name changes and organisation, 
Specifically removing the `scraped_`, you can check the [official jagex names here](https://gameval.z-kris.com/#game=runescape&type=loc&cache=2670&q=1177&searchMode=substring&searchBehavior=filter), however they are not always great or consistent so best to use them as a guide.
As for organising just moving them to vaguely correct area based .obj.toml files, e.g. magical wheat is probably specific to just Zanaris, mineral veins from living rock caverns etc... rather than one big dump.obj.toml

PR #1295, opened 2026-09-13
Restored functionality on all deposit boxes.

PrettyPrincessPotato:
There is a small caveat, I haven't been able to properly identify which boxes are where, and instead of painstakingly tracking them down individually I added their ID to the examine text to help identify where they should go.

This started as an individual fix to the Ardougne deposit box, but since I could see the rest of the boxes that didn't have a string name attached to it, I thought it would be appropriate to at least restore functionality.

### DIFF
--- a/data/area/kandarin/ardougne/east_ardougne.objs.toml
+++ b/data/area/kandarin/ardougne/east_ardougne.objs.toml
@@ -216,3 +216,31 @@ examine = "Handy for boarding the ship."
 [captain_barnaby_ship_ladder]
 id = 9745
 examine = "A sturdy wooden ladder."
+
+[bank_deposit_box_east_ardougne]
+id = 34755
+examine = "It lets you put the items into your bank."
+
+# I'm identifying each of the remaining deposit boxes, the final number representing their string ID.
+# I'm adding the string ID to the examine text to let us know that it's a lost deposit box, and needs to be properly assigned
+# To an area.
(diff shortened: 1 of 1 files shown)


data/area/kandarin/ardougne/east_ardougne.objs.toml
```diff
@@ -216,3 +216,30 @@ examine = "Handy for boarding the ship."
...
 id = 9745
 examine = "A sturdy wooden ladder."
+
+[bank_deposit_box_east_ardougne]
+id = 34755
+examine = "It lets you put the items into your bank."
+
+# I'm identifying each of the remaining deposit boxes, the final number representing their string ID.
+# I'm adding the string ID to the examine text to let us know that it's a lost deposit box, and needs to be properly assigned to an area.
+# This allows all deposit boxes to work functionally, even if they're not all properly identified.
```
GregHib: See: ObjectUsageFinder.kt to identify where they are in the world

GregHib (changes requested):

data/area/kandarin/ardougne/east_ardougne.objs.toml
GregHib: Alternatively we can also look at the official names
https://gameval.z-kris.com/#game=runescape&type=loc&cache=2670&q=%5E%286836%7C9398%7C20228%7C25937%7C39830%29%24&searchMode=regex&searchBehavior=filter

GregHib (approved):

PR #1300, opened 2026-09-13
Add in-client account registration

HarleyGilpin:
## Summary

Adds account creation from the client's login screen ("New user").

## Changes

- **network**: `RegistrationServer` dispatched from `GameServer` for opcodes 28/22, `RegistrationResponse` codes, `RegistrationValidator`, per-IP `RegistrationLimiter`, `AccountCreator` interface. Login (`LoginServer`/`PasswordManager`) accepts email-shaped account names beyond 12 characters via `AccountNames`. Invalid names now get response 3 ("Invalid username or password") instead of 11, which the client shows as the "password is an extremely common choice" warning (the message behind #1307).
- **engine**: `PlayerAccountLoader` also implements `AccountCreator`: it reserves the email in `AccountDefinitions` on the game thread and persists via `Storage.create`. `AccountManager.create` gives email accounts a placeholder display name from the local part (numbered if taken) and flags `choose_name`, so login-created and registered email accounts behave the same. `Storage.create` persists a brand-new account atomically (`FileStorage`: `createNewFile` + temp file + atomic move; `DatabaseStorage`: single transaction, unique-violation → `false`). `AccountDefinitions.remove` for rollback. `DisplayNames` holds the name rules and suggestion generator.
- **database**: `accounts.name` and `abuse_reports.reporter` widened to 254 with idempotent migrations in `connect`.
- **game**: at the end of character creation, registered accounts get the authentic **Character Name** panel (interface 1028 components 154/158-166): the server runs client script 3943 to show it, sends six suggestions as client strings 337-342 and runs 3952; Continue / a suggestion runs 3945 which returns the typed name as a string entry. Taken or invalid names re-prompt with a message and fresh suggestions ("More"/"Previous Suggestions" page through them). When character creation is disabled a name-entry dialogue is used instead. `rename`/`nameTaken` extracted from `NameChange`; the `::rename` command now rejects taken names too. New settings:
  ```
  accounts.registration=true
  accounts.registration.maxPerIP=3
  accounts.registration.windowMinutes=60
  ```
  `development.accountCreation` (auto-create on unknown login) is unchanged.

## Testing

- Unit tests for the server decode paths (hand-built RSA/XTEA packets), validator, limiter, `GameServer` dispatch, login with email names, registration via `PlayerAccountLoader`, `DisplayNames` (rules + suggestions), `AccountDefinitions.remove`, `Storage.create` on both backends, and `WorldTest`s for the name panel (confirm, taken/invalid/accepted names, suggestion paging and picking) and the dialogue fallback.
- End-to-end against a running server with raw sockets: every reply code above, duplicate/rate-limited creates, then login with the wrong and the correct password.


skalfate:
Amazing stuff!! Some parts I can understand and some not but it doesn't matter to me. Its awesome. This is very appreciated. ❤️

HarleyGilpin:
I am going to do some work off this branch to add Ktor restful web API to enable web registration, player counts, and adventure logs. So I'm converting this PR into a draft.

database/src/main/kotlin/world/gregs/voidps/storage/DatabaseStorage.kt
```diff
@@ -218,6 +219,39 @@ class DatabaseStorage : Storage {
...
+                    .select(AccountsTable.id, AccountsTable.name)
+                    .where { LowerCase(AccountsTable.name) eq lower }
+                    .associate { it[AccountsTable.name].lowercase() to it[AccountsTable.id] }
+                saveExperience(list, playerIds)
+                saveLevels(list, playerIds)
+                saveVariables(list, playerIds)
+                saveInventories(list, playerIds)
+                saveOffers(list, playerIds)
+                saveHistories(list, playerIds)
+                true
```
GregHib: This'll need updating with the latest stats sections

engine/src/main/kotlin/world/gregs/voidps/engine/client/PlayerAccountCreator.kt
```diff
@@ -0,0 +1,96 @@
...
+import world.gregs.voidps.network.login.Registration
+import world.gregs.voidps.network.login.registration.RegistrationResponse
+import world.gregs.voidps.network.login.registration.RegistrationValidator
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Creates accounts registered from the client's login screen.
+ * The email address is the account name; a display name is derived from it until the player chooses their own on first login.
+ */
+class PlayerAccountCreator(
```
GregHib: I don't follow why this is all necessary? There are already uniqueness checks for usernames, which are case insensitive and we can just check if the username is an email we assign an incrementing player1/2/3 as the displayName and rename it once they're in-game and have picked a name?

database/src/main/kotlin/world/gregs/voidps/storage/DatabaseStorage.kt
HarleyGilpin: Extracted a shared `saveSections` used by both `save` and `create` so the two can't drift again, a1b40205c11e3e2ed56eec62f898b235e3add976

engine/src/main/kotlin/world/gregs/voidps/engine/client/PlayerAccountCreator.kt
HarleyGilpin: Fair point, simplified in eb403ad011ffa4b4e563579818120620e6d15c79:

- Placeholder name is now the next free `Player1`/`Player2`/… via the existing case-insensitive `AccountDefinitions.get`; `DisplayNames.unique` and the email local-part derivation are gone.
- Dropped the in-flight set and the extra `storage.exists` lookup. The availability check and `accountDefinitions.add` now run together on the game thread, with `Storage.create` returning `false` on a duplicate as the backstop.
- Removed the unread `registered` variable.

The rename happens on first login as before: `choose_name` opens the Character Name panel at the end of character creation.

engine/src/main/kotlin/world/gregs/voidps/engine/EngineModules.kt
```diff
@@ -36,7 +37,8 @@ fun engineModule(files: ConfigFiles) = module {
     single { AccountManager(get(), get(), get(), AppearanceOverrides().apply { load() }) }
     single { AccountDefinitionsReloader(get(), get(), get()) }
     // IO
-    single { PlayerAccountLoader(get(), get(), get(), get(), get(), Contexts.Game) }
+    single { PlayerAccountLoader(get(), get(), get(), get(), get(), Contexts.Game, get()) }
+    single { PlayerAccountCreator(get(), get(), get(), gameContext = Contexts.Game) }
```
GregHib: These do the same thing?

HarleyGilpin:
8447e46ac2ace2208d711088ba723234ddceb330 fixes the "Your password is an extremely common choice" message from #1307: `PasswordManager` returned response 11 for names over 12 characters (and now malformed emails), and 11 is the client's weak-password slot. It sends 3 ("Invalid username or password") now, and the unreachable duplicate name check in `LoginServer` is gone. Pre-existing on main, so happy to split it out if you'd rather.

engine/src/main/kotlin/world/gregs/voidps/engine/client/PlayerAccountCreator.kt
HarleyGilpin: Went back to deriving the placeholder from the email in 1320c3f02e575c506f14d114e52002f2e3c5dd18: since the name panel suggestions build on the current name, "Bob smith" gives better suggestions than "Player7". Still the existing case-insensitive checks plus a numeric suffix if taken; the in-flight set and extra storage lookup stay gone.

HarleyGilpin:
Two more from testing:

- 06860daf4 - name panel suggestions were always one call behind (the previous page, or the previous player registered on the same client). The 634 client queues varcstr updates and applies them after it has read all packets in the frame, but runs script packets immediately, so 3952 read the old strings. The render script is now queued a tick after the strings.
- 8181bbc37 - logging in while already online on an account created since startup (`development.accountCreation`, not yet saved) answered 4 "account disabled" because `PasswordManager` treated a missing file as disabled before the online check ran. The in-memory hash is checked instead, so it falls through to the normal already-online reply. Pre-existing on main.

engine/src/main/kotlin/world/gregs/voidps/engine/EngineModules.kt
HarleyGilpin: Yeah, merged them in d93879b3e. `PlayerAccountCreator` is gone: `PlayerAccountLoader` implements `AccountCreator` (reserve the email in `AccountDefinitions` on the game thread, `Storage.create`, roll back on failure) and `AccountManager.create` sets the placeholder display name for email accounts, so the login path with `development.accountCreation` and the registration path share it.

GregHib:
It doesn't seem to work? I fill out the details and press enter and the button gets disabled but nothing happens
image

HarleyGilpin:
Cache patch was required to fix the issue. see updated cache files here: https://dl.00002011.xyz/cache.zip

engine/src/main/kotlin/world/gregs/voidps/engine/data/file/FileStorage.kt
```diff
@@ -332,6 +334,24 @@ class FileStorage(
...
+        }
+        val temp = directory.resolve("${account.name.lowercase()}.toml.tmp")
+        try {
+            account.save(temp)
+            Files.move(temp.toPath(), file.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)
+        } catch (e: Exception) {
+            temp.delete()
+            file.delete()
+            throw e
+        }
```
GregHib, line 352: Shouldn't the temp be in finally so it's cleared up after the move is complete?

game/src/main/kotlin/content/entity/player/DisplayNameSelection.kt
```diff
@@ -0,0 +1,31 @@
...
+import content.social.friend.nameTaken
+import content.social.friend.rename
+import world.gregs.voidps.engine.data.definition.DisplayNames
+import world.gregs.voidps.engine.entity.character.player.Player
+import world.gregs.voidps.engine.entity.character.player.name
+
+/**
+ * Accounts registered with an email address are given a placeholder display name until the player picks one on first login
+ */
+suspend fun Player.chooseDisplayName() {
```
GregHib: This can be moved inside of Introduction and the tests can access it by createPlayer/playerSpawn

GregHib (changes requested):

engine/src/main/kotlin/world/gregs/voidps/engine/data/file/FileStorage.kt
HarleyGilpin: Yes, moved to a finally in 9d7c96993aa362daf662df9fdeb93801956f7820 (the atomic move already consumes it on success, so the delete is a no-op there and cleanup on failure).

game/src/main/kotlin/content/entity/player/DisplayNameSelection.kt
HarleyGilpin: Moved into Introduction as a private suspend function in 6bd9f5b37e70555b0e9abf34abe48d516edf1faf; DisplayNameSelection.kt is gone and the test (now IntroductionTest) goes through createPlayer with world.start.creation off so playerSpawn drives the prompt.

PR #1308, opened 2026-09-16
Add Wine of zamorak spawn to the Asgarnia Chaos Temple

HarleyGilpin:
## Summary

Adds the Wine of zamorak floor spawn to the Chaos Temple north-west of Goblin Village (`2931,3515`, 50 tick respawn) with the 2011 theft mechanic, and fixes a few things found on the way.

**Wine of zamorak** (`ChaosTempleWine`, hooked through `takeable`)
- Taking it by hand from the spawn tile: "STOP STEALING MY WINE! GAH!", Flames of Zamorak gfx, 30-40 damage, Attack/Strength/Defence/Ranged/Magic drained by 1 per 13 current levels (max 3), and every live Monk of Zamorak in the temple attacks.
- Telekinetic Grab, a wine on any other tile, or a full inventory: no punishment.
- Arriving after another player has already taken it: no punishment, just the usual "Too late".

**Telekinetic Grab**
- Never passed `telegrab = true` to `takeable` handlers, so content could not tell a grab from a hand pickup (the `ZogreFleshEaters` telegrab branch was unreachable). Fixed.
- Projectile and impact now end at the item's height when it sits on a table. Object definitions can declare `height` in client units (the temple table is 384); `FloorItem.height()` reads it. Projectile heights are divided by 16 to match the client, and `areaGfx` clamps to the packet's byte range.
- Flight time halved and the item is removed on the tick the impact plays instead of six ticks after casting.

**Monks of Zamorak**
- `height = 40` on the base and Goblin Village defs (Ourania already had it) and `height = 50` on the zap gfx so it plays at hand height rather than the feet.

## Verification

- `ChaosTempleWineTest`: hand pickup, telegrab safe, off-tile harmless, respawn, telegrab race, late-arrival race.
- `TelekineticGrabTest`, `FloorItemRespawnTest`, `TakeTest`, `MageArenaTest` still pass.
- Checked in-game: projectile lands on the wine, monks aggro, stats drain, wine respawns.

Sources: 2011 wiki revision of Wine of zamorak for the numbers and messages.


data/skill/magic/spellbook/spellbook.gfx.toml
```diff
@@ -11,7 +11,7 @@ id = 143
 height = 10
 delay = 48
 curve = 16
-time_offset = 42
+time_offset = 21
```
GregHib, line 14: What's this value based off of?

game/src/main/kotlin/content/entity/item/FloorItemHeight.kt
```diff
@@ -0,0 +1,13 @@
...
+import world.gregs.voidps.engine.entity.obj.GameObjects
+import world.gregs.voidps.engine.entity.obj.ObjectLayer
+
+/**
+ * Height the item is displayed at, in client units (one tile is 512 wide). The client raises item piles
+ * by the model height of the object under them (tables, altars), so projectiles and graphics aimed at
+ * the item need the same offset. Object definitions declare it with a `height` param; model heights of
+ * pre-version-13 models are multiplied by 4 by the client.
+ */
+fun FloorItem.height(): Int = GameObjects.getLayer(tile, ObjectLayer.GROUND)?.def?.get("height", 0) ?: 0
```
GregHib, line 13: This assumes that the item is always spawned on the same height tile, which isn't probably rarely the case. We should use a proper fix (common on other class).

game/src/main/kotlin/content/skill/magic/book/modern/TelekineticGrab.kt
```diff
@@ -52,17 +54,20 @@ class TelekineticGrab : Script {
             sound("tele_grab_cast")
             exp(Skill.Magic, Tables.int("spells.$spell.xp") / 10.0)
 
-            val clientTicks = shoot("tele_grab_travel", floorItem.tile)
+            val height = floorItem.height()
```
GregHib, line 57: I think a better fix for this will be to check if the tile is blocked, and if it is assume it's a table and that the item will be at the fixed table height. We won't be able to know the real height without storing and checking tile heights which seems overkill for this one use-case.

GregHib (changes requested):

PR #1317, opened 2026-09-17
Rebuild the cached object map when its settings change

### LINKED ISSUE #1314: map_objects.dat cache is not invalidated when object storage settings change
## Description

Which map objects are stored server-side is decided once and then cached to disk, and the cache is not invalidated when the setting that governs it changes.

```kotlin
// engine/src/main/kotlin/world/gregs/voidps/engine/data/definition/MapDefinitions.kt:43-47
if (objectsFile.exists() && collisionsFile.exists() && !configFiles.cacheUpdate) {
    val zones = collisions.load(collisionsFile)
    GameObjects.load(objectsFile)
    ...
```

`configFiles.cacheUpdate` is derived from cache-file timestamps only (`ConfigFiles.kt:46`, `cacheChanged(lastUpdated, cachePath)`). It knows nothing about `game.properties`.

Meanwhile the decision being cached is:

```kotlin
// engine/src/main/kotlin/world/gregs/voidps/engine/entity/obj/GameObjects.kt:127-133
fun set(id, x, y, level, shape, rotation, definition) {
    collisionAdd.modify(definition, x, y, level, shape, rotation)
    if (interactive(definition)) { map[...] = value(false, id, shape, rotation); size++ }
}

var storeUnused = Settings["development.loadAllObjects", false]
private fun interactive(definition) = storeUnused || definition.options != null || definition.contains("id")
```

So flipping `development.loadAllObjects` has no effect until `data/.temp/map_objects.dat` is deleted or the game cache itself changes. The `.dat`, not the property, decides what is interactable — and a `.dat` written months ago under the opposite setting is used silently.

## Impact

Low but confusing to debug: the symptom is content that behaves as though the flag were still at its old value, with nothing in the logs to indicate why. If the stale `.dat` was written with `storeUnused = false`, `GameObjects.find(tile, ...)` throws `error("Object not found at $tile")` (`GameObjects.kt:226,241,251`), `remove()` on an unstored original silently no-ops, and `replace()` adds the replacement without ever removing the original's collision.

## Suggested fix

Include the object-storage settings in the cache-validity check — e.g. write `storeUnused` into the `.dat` header and reject the file when it differs, or fold the relevant settings into whatever `cacheUpdate` hashes.

## Related: `interactive()`'s `contains("id")` is always false

`definition.contains("id")` can never be true, so `interactive()` is effectively `storeUnused || definition.options != null`:

- `Parameterized.contains(key: String?)` resolves the name through `Params.idOrNull(key)` (`cache/.../definition/Parameterized.kt:25-31`), and `Params.idOrNull` has no `"id"` branch.
- It could not have one, because `ObjectDefinitions.load` consumes `id` as the numeric definition id before it ever reaches the params map (`engine/.../definition/ObjectDefinitions.kt:58`).

While there: that same `contains` has `val key = Params.idOrNull(key) ?: false`, which widens an `Int?` to `Any` and then calls `containsKey(false)`. It returns false either way, so the behaviour is right by accident; `?: return false` is what was meant.

HarleyGilpin:
Fixes #1314.

`MapDefinitions.load` reads `data/.temp/map_objects.dat` verbatim whenever it exists and `configFiles.cacheUpdate` is false. That flag is derived from the game cache's timestamps only (`ConfigFiles.kt:46`) and knows nothing about `game.properties`.

The thing being cached is `interactive()`'s verdict, which depends on `development.loadAllObjects`. So changing that setting did nothing until someone deleted the `.dat` by hand — the file, not the property, decided what was interactable, and a `.dat` written months ago under the opposite setting was used in silence. If it happened to be one written with `storeUnused = false`, the symptoms are content-side and unhelpful: `GameObjects.find` throwing `Object not found at $tile`, `remove()` silently no-oping on an unstored original, and `replace()` adding the replacement without ever clearing the original's collision.

## Changes

The file now opens with a magic and a flags int recording the `storeUnused` it was written with. `load` returns `INVALID` when either doesn't match and leaves the map untouched, so `MapDefinitions` falls through to decoding the cache and rewriting the file — self-correcting from then on.

Existing `.dat` files have no header, so they're rejected and rebuilt once. That's deliberate: their first int is an object count, which can't reach the magic, so there's no ambiguity. The cost is one slower startup.

Objects are now loaded before collision, since there's no point loading collision for a map that's about to be re-decoded.

Also drops the unreachable `contains("id")` arm of `interactive()`. `Parameterized.contains(String)` resolves through `Params.idOrNull`, which has no `"id"` entry and can't have one — `ObjectDefinitions.load` consumes `id` as the numeric definition id before it could ever reach the params map. The check was always false, so this is behaviour-preserving.

## Tests

Three cases in `GameObjectHashMapTest`: a save/load round trip, rejection of a file written with a different `storeUnused`, and rejection of a headerless file in the old format.

## Noted but not changed

`Parameterized.contains` has `val key = Params.idOrNull(key) ?: false`, which widens an `Int?` to `Any` and then calls `containsKey(false)`. It returns false either way so the behaviour is right by accident; `?: return false` is what was meant. Left alone as it's in a different module and unrelated to this fix.


GregHib (commented):
If there is a reload and this field has changed can't we just delete the file and rebuild?

HarleyGilpin:
That is what happens, rejecting the file falls through to the existing `else` branch, which decodes from the cache and rewrites it. Deleting it first would be the same thing minus one write.

The part that needs somewhere to live is knowing the field changed. `cacheUpdate` is derived purely from cache file timestamps, so on its own nothing in the process has any memory of what `storeUnused` was when the `.dat` was written. That's the only reason I put it in the file.

If you'd rather leave the format alone, the cheap version is to fold `game.properties` into the staleness check, `cacheChanged` already compares mtimes against `modified.dat`, so it'd be one more path, and `cacheUpdate` has exactly one consumer (`MapDefinitions`) so nothing else gets invalidated by it. Two caveats: any edit to `game.properties` would rebuild the map cache, not just edits to this field, and it wouldn't cover the case where `./game.properties` is absent and the bundled resource is used, since there's no file to stat.

GregHib:
Hmm yeah, I'm pretty sure at the time I knew it's the only properly that couldn't be live reloaded cause of this but I was fine deleting the file and restarting. Live reload seems overkill but also changing the setting and restarting and it not applying isn't acceptable either.

Best options seem to be:
1. Store the `storeUnused` flag in modified.dat and compare it to the game.properties one to check for reload
2. Some kind of timestamp check on the file, if it's older than todays date delete it, otherwise write the new format
3. Store the unused dat under a different name (but duplicate data left over)

GregHib (changes requested):

PR #1318, opened 2026-09-17
Stop leaking instance regions, and survive a bad tick

### LINKED ISSUE #1311: Instances region pool leaks; exhaustion throws out of the game loop and stops ticks
## Description

`Instances` hands out regions from two fixed pools and relies on `free()` to return them. Several paths never free, so the pools drain monotonically, and exhaustion throws in a place that terminates the game loop.

```kotlin
// engine/src/main/kotlin/world/gregs/voidps/engine/map/instance/Instances.kt:19-23
fun small(): Region {
    val region = small.pollFirst()   // null once the deque is empty
    used.add(region)
    return region                    // non-null Region — unboxing throws NPE
}
```

Pool sizes from `reset()` are 1377 small and 700 large. `reset()` is only called from the `init` block and from tests, never at runtime.

The existing test already documents the failure:

```kotlin
// engine/src/test/kotlin/world/gregs/voidps/engine/map/instance/InstancesTest.kt:56-64
@Test
fun `No more instances throws exception`() {
    repeat(1377) { Instances.small() }
    assertThrows<NullPointerException> { Instances.small() }
}
```

## Leaks

`Instances.free()` has exactly one caller, `Player.clearInstance()` (`game/src/main/kotlin/content/quest/Cutscene.kt:165`). Regions escape it in several ways:

- **Dungeoneering never frees on logout.** `largeInstance()` (`Cutscene.kt:119-123`) never sets `instance_logout`, and `DungeonEntrance.kt:210-211` explicitly uses `smallInstance(logout = false)`, so the `playerDespawn` handler in `InstanceLogout.kt:17` does nothing. Every party that ends by logout or disconnect leaks its region permanently.
- **Overwriting.** `smallInstance()` allocates and overwrites `"instance"` unconditionally, so any path that starts a second instance before clearing the first orphans the previous region in `used` forever.
- **Death.** Nothing frees an instance on death except the Fight Cave's own handler (`TzhaarFightCave.kt:210-219`).
- **Party aliasing.** All members share one id (`DungeonEntrance.kt:217`). The first member to call `clearInstance` frees the region back to the pool while the others are still inside it, so it can be handed to a second party while in use.

`"instance"` is also persisted to the save file (`PlayerSave.kt:411` serialises `variables.data`) while `Instances.used` is not, so a stale id outlives a restart.

## Impact

On exhaustion, `used.add(null)` succeeds and then unboxing the `@JvmInline value class Region` throws NPE. Where that lands decides the severity:

- From an instruction it is swallowed (`InstructionTask.kt:21` catches `Throwable`), leaving the player half-entered.
- From a queue, timer or script stage it escapes `tick()` into `GameLoop.start`:

```kotlin
// engine/src/main/kotlin/world/gregs/voidps/engine/GameLoop.kt:15-33
try {
    while (isActive) {
        ...
        tick()
        ...
    }
} catch (e: Exception) {
    if (e !is CancellationException) {
        logger.error(e) { "Error in game loop!" }
    }
}
```

The `catch` is outside the `while`, so a single NPE ends the loop. The process stays up with no ticks — no movement, no combat, no saves.

## Suggested fix

- Free on every exit path, including death and disconnect, and drop `logout = false` where it is not needed.
- Refcount or owner-check party instances so one member leaving cannot return a region others are still inside.
- Have `small()`/`large()` return `Region?` and let callers fail gracefully, rather than NPE-ing out of the game loop.
- Move the `try`/`catch` in `GameLoop.start` inside the `while` so one bad tick cannot kill the server, and consider re-validating `"instance"` against `Instances.isInstance` on load, since it is persisted.

HarleyGilpin:
Fixes #1311.

Two halves of the same failure: the instance pool only ever drains, and running out of it stopped the server.

## The leaks

`Instances.free` has exactly one caller, `Player.clearInstance`, and several paths never reach it:

- **Dungeoneering frees inside a `longQueue`** (`DungeoneeringParty.kt:375`), which a disconnecting player never runs — and only `if (last)`, where `last = player == leader && dungeonMembers.size == 1`. So if the leader drops out first, no later leaver can ever satisfy that condition and the region is stranded permanently.
- **Nothing released the region when a player with `instance_logout` off despawned.** That flag was doing double duty: "teleport me out on logout" and, incidentally, "hand the region back".
- **Party members share one id** (`DungeonEntrance.kt:217`), so the first to call `clearInstance` returned a region the rest were still standing in.

Regions are now refcounted. Allocating counts one holder, `claim()` adds another, and `free()` reports whether the last holder has let go — so `clearInstance` only tears the map down for the final leaver, and an early leaver can't pull the floor out from under their party. Dungeoneering members each claim on entry and release on the way out, in any order, and a despawning player hands their hold back even when the content runs its own exit.

The two despawn paths compose: whichever of `InstanceLogout` and `DungeoneeringParty` runs first, each player releases exactly once, because `clearInstance` removes `"instance"` and returns early on a second call.

## Survivability

`GameLoop.start` had its `try`/`catch` outside the `while`, so a single exception escaping `tick()` ended the loop. The process stayed up with no ticks at all — no movement, no combat, no saves — which is a worse failure than crashing.

An exhausted pool did exactly that: `Instances.small()` returns a non-null `Region`, `pollFirst()` returns null once the deque is empty, and unboxing the `@JvmInline value class` throws. From a queue, timer or script stage that reached `GameLoop`. It's caught per tick now, and `CancellationException` still propagates so shutdown is unaffected.

## Tests

Three cases in `InstancesTest` covering claim/free refcounting, freeing something nobody holds, and claiming something nobody holds. The existing `No more instances throws exception` still passes — the nullable-allocation refactor would have touched all 17 `smallInstance`/`largeInstance` call sites for little gain now that a bad tick is contained, so I left the signature alone; happy to do it separately if you'd rather.

`GameLoopTest`'s `Exception in stage` asserted the old behaviour (`assertFalse(job.isActive)`), so it's inverted to assert the loop carries on. Note that class is `@Disabled("Flaky")` at the class level, so it isn't running in CI either way.

Full `:engine:test` passes.

## Conflict note

Touches `clearInstance` in the same place as #1312, which adds an `Instances.reserved` guard a few lines up. Whichever lands first, the other is a trivial rebase — the two guards are complementary, not alternatives.


GregHib:
> Dungeoneering frees inside a longQueue (DungeoneeringParty.kt:375), which a disconnecting player never runs

A long queue should be run on disconnect/logout, that might be more the issue here

GregHib (changes requested):
^ needs investigating

PR #1319, opened 2026-09-17
Add account REST api to the webserver

HarleyGilpin:
Depends on #1300 (account registration); the first commits here are that branch and will drop out once it merges.

## What

Adds an authenticated JSON api under `/api` on the existing `:web` ktor server so a website or launcher can manage accounts on the running server. Today an external writer (e.g. a website inserting into the database) is invisible to login until `::reload accounts` or `storage.accounts.reloadMinutes` fires; accounts created through the api are live immediately and the server keeps ownership of defaults, hashing and validation.

Off by default. Settings (documented in `game.properties` and the README):

```
web.api.enabled=false
web.api.token=
web.api.registration=true
web.api.registration.maxPerIP=3
web.api.registration.windowMinutes=60
```

| Method | Path | Body | Success |
| --- | --- | --- | --- |
| `GET` | `/api/status` | | name, world, revision, players, uptime |
| `GET` | `/api/accounts/{name}` | | account by account or display name |
| `POST` | `/api/accounts` | `{name, password, displayName?}` | 201 |
| `PUT` | `/api/accounts/{name}/password` | `{password}` | 204 |
| `PUT` | `/api/accounts/{name}/name` | `{displayName}` | 200 |

Every request needs `Authorization: Bearer <web.api.token>`; errors are `{"error","message"}` with 400/401/403/404/409/429/503. `name` is a 1-12 character username (also the display name) or an email address, in which case `displayName` is required. Underscores and dashes are treated as spaces, matching the client's name entry. Creation is rate limited per end-user address taken from `X-Forwarded-For`, which is only trusted after the bearer check.

## How

- `:web` owns the contract (`AccountService`, result enum, DTOs) and the routes; `game` wires `WebAccountService` onto the engine, the same shape as `RegistrationServer` ↔ `AccountCreator`.
- `PlayerAccountCreator.create(accountName, hash, displayName?, hostname)` handles username and email names; the in-client path delegates to it unchanged.
- New `PlayerAccountUpdater`: online players are changed in memory and queued for saving; offline accounts are loaded, changed and saved on IO, then the in-memory definition is updated on the game thread. Returns `BUSY` while a save is in flight.
- `Player.rename` / `nameTaken` and `Settings.world` / `worldName` move from game content into the engine so the updater can notify friends.
- `AccountDefinitions.update` now re-keys the owner's clan entry (previously orphaned on every rename).
- The webserver starts after content load (it needs engine services) and the webclient zip is optional when only the api is enabled.
- JSON via `kotlinx-serialization` on `:web` only (compiler plugin applied the same way as `composeCompiler` in `tools/app`); auth via `ktor-server-auth` bearer with a constant-time compare.

## Tests

- `ApiRoutesTest` (ktor `testApplication`, stub service): auth, every error mapping, forwarded address, malformed JSON, routes absent when disabled.
- `PlayerAccountCreatorTest`, new `PlayerAccountUpdaterTest`, `AccountDefinitionsTest` clan re-key, `WebAccountServiceTest` (validation, limiter, code mapping).
- Full `:engine`, `:network`, `:web` and `:game` suites pass; `-PincludeDb` builds.
- Smoke tested against a running server with curl: create (both name styles), lookup, password change (hash verified), rename with `name_history`, 401/400/409/429/404 paths, webclient still served.


HarleyGilpin:
@GregHib - This PR is ready for review. It depends on Add in-client account registration- #1300, please review and merge PR 1300 first before we merge this PR.

PR #1320, opened 2026-09-17
Add cache auto-downloader at startup

HarleyGilpin:
## Summary

Adds a startup step that downloads and installs the game cache when it is missing or out of date, so a fresh checkout or release bundle can start without manually extracting `cache.zip` into `data/cache/`.

- New `CacheDownloader` in the `cache` module (JDK only: `HttpClient`, `ZipInputStream`, `MessageDigest`, `HexFormat`, `JOptionPane`; no new dependencies).
- `Main.kt` calls `CacheDownloader.ensure(settings)` before `Cache.load` and returns if the cache could not be installed.
- New property `storage.cache.download.url` (default `https://dl.00002011.xyz`, blank disables). The host serves `version.json` (`{"file": "cache.zip", "sha256": "<hex>"}`) and the zip.
- README setup steps updated; manual extraction remains as the fallback.

## Behaviour

| cache files | marker | remote | action |
|---|---|---|---|
| missing | – | ok | prompt to install (auto-yes when headless); decline exits |
| missing | – | unreachable | error, server does not start |
| present | none | ok | record remote sha as installed, no download |
| present | equal | ok | nothing |
| present | differs | ok | prompt to update; decline keeps the existing cache |
| present | any | unreachable | warning, continue with existing cache |

The zip is hashed while streaming into `cache.zip.part`; on mismatch it is deleted and nothing else is touched. `main_file_cache.*` entries are extracted regardless of folder prefix (basename only, so zip-slip safe), staged as `.part` files, checked for `dat2` + `idx255`, then atomically moved into place. The installed sha is written to `data/cache/cache.sha256` (gitignored). On-disk cache files are never hashed because tooling modifies them after install.

## Notes

- The sha256 comes from the same host as the zip, so it guards against corruption, truncation and stale copies, not a compromised host.
- Docker: images built with `COPY ./data/` have no marker, so a fresh container does one GET and adopts the remote sha. Set `storage.cache.download.url` to blank via `environment:` for offline containers.

## Testing

- `CacheDownloaderTest`: 19 unit tests (install, folder-prefixed zips, sha mismatch, missing required files, adopt, up to date, update, declines, unreachable, malformed json, stale `.part` cleanup, truncated download, invalid file name, version parsing).
- End to end against a local `python3 -m http.server` hosting a `cache/`-prefixed zip of the real cache: fresh install, up to date, tampered marker, deleted marker, unreachable with and without files, all headless via `:game:run`.
- `spotlessCheck` and `SettingsTest` pass.


cache/src/main/kotlin/world/gregs/voidps/cache/CacheDownloader.kt
```diff
@@ -0,0 +1,310 @@
...
+        const val VERSION_FILE = "version.json"
+        private const val STAGED_SUFFIX = ".part"
+        private const val CACHE_FILE_PREFIX = "${FileCache.CACHE_FILE_NAME}."
+        private val REQUIRED_FILES = listOf("${FileCache.CACHE_FILE_NAME}.dat2", "${FileCache.CACHE_FILE_NAME}.idx255")
+        private const val BUFFER_SIZE = 64 * 1024
+        private const val UNKNOWN_LENGTH_STEP = 25L * 1024 * 1024
+        private val FILE_NAME_PATTERN = Regex("[A-Za-z0-9_-][A-Za-z0-9._-]*")
+        private val SHA_PATTERN = Regex("[0-9a-f]{64}")
+        private val FILE_KEY = Regex("\"file\"\\s*:\\s*\"([^\"]+)\"")
+        private val SHA_KEY = Regex("\"sha256\"\\s*:\\s*\"([0-9a-fA-F]{64})\"")
```
GregHib, line 279: This seem unnecessary and add extra overhead

GregHib (changes requested):
The overhead of this is quite high, checking takes ~480ms on my machine which is +10% increase on startup

PR #1321, opened 2026-09-18
Add a server console for operator commands

HarleyGilpin:
Adds a console to the terminal running the server, so an operator can type commands rather than only read log output.

```
help                    List the available console commands
clear                   Clear the console output
players                 Count and list the online players
kick (player-name)      Disconnect a player
announce (message)      Send a message to all online players
save                    Save all online players
shutdown [seconds]      Shutdown the server, optionally after a countdown
```

Arguments with spaces are quoted, e.g. `announce "Server restarting soon"`. `console.enabled=false` in `game.properties` turns the whole thing off.

### Console commands vs player commands

Player commands take a `Player` to reply to and to check rights against; the console has neither. Rather than fake a player, the dispatch table is separate and its handlers delegate to the same functions - `saveAll()` and `queueShutdown()` are extracted from the existing `save` and `update` commands for exactly that, and `kick` goes through `AccountManager.logout`.

Lines are read on a daemon thread but executed by `ConsoleCommands` as a game loop stage, so handlers only touch game state from the game thread. Replies are written with a coloured gutter so they read apart from the server's own logging, and each command is recorded in the audit log:

```
› players
│ 1 player online:
│   Pixel Tile(3222, 3222, 0)
INFO [SaveLogs] Saved 4 logs to disk in 1ms.
```

### Terminal handling

Given a real terminal the input line stays pinned below the log, log lines print above whatever is half-typed, with tab completion (command names, and argument values from the existing `CommandArgument.autofill`, so `kick` completes online players), history on up/down, and the usual editing keys.

`stty` puts the terminal into character-at-a-time mode. Signals are deliberately left enabled rather than using full raw mode, so **Ctrl + C still interrupts the jvm and runs its shutdown hooks** as the readme documents, and `run-server.sh`'s `trap cleanup INT` keeps working. Saved terminal settings are restored on every exit path and from a shutdown hook.

### Tests

61 tests, none of which need a terminal, a `FakeTerminal` feeds bytes in and the rendered bytes are asserted, covering key decoding, editing, history, completion, the print-above redraw including the wrapped-line case, multi-byte input, EOF and the plain fallback.

Also exercised against a running server under a pty: prompt and logs interleaving, tab completion, history, `clear`, window resize, piped stdin, and Ctrl + C leaving the shell with `icanon echo` intact.

---

### Since the first push

- **Unquoted arguments** - a command can mark its last argument as taking the rest of the line, so `announce server restarting soon` and `kick harley gilpin` work without quotes. Quoting still works, and commands without one still report extra arguments.
- **Suggestions and aliases** - an unknown command suggests the nearest by `Distance.levenshtein` the way `Commands.find` does instead of dumping the whole list; `quit`/`exit`/`stop`, `list`/`online` and `?` are aliases.
- **`help <command>`** describes one command; bare `help` still lists everything.
- **Editing keys** - Ctrl+A/E, Ctrl+U/K and Ctrl+W.
- **New commands** - `reload <config-type>` (the player command used its `Player` in two of twenty-five branches, so the rest is now a shared `reloadConfig()`; both sides report an unknown type rather than doing nothing silently), `mute`/`unmute` delegating to the same `Player.mute`/`unmute` the moderation commands use, and `uptime` (jvm uptime, ticks, players, npcs, heap - no new engine state).
- **stderr is wrapped too.** Logging goes through stdout, but `printStackTrace` writes straight to `System.err` in six places on the server path and nothing installs a default uncaught exception handler, so a stack trace from another thread printed over the input line.
- **Piped output no longer gets drawn into.** `stty` only probes the input side, so `run-server.sh | tee log` wrote escape sequences into the pipe; taking the terminal now also requires `System.console() != null`.
- **Completion no longer reads live game state.** `kick`'s autofill iterated `Players` from the console thread while the game thread mutated it; names come from the account definitions the player commands autofill from, and an autofill that throws costs the completion rather than the key press. History is also capped.

Tests are up to 100 across the console, including the game-side integration tests for announce, save, shutdown, mute, uptime, aliases and unquoted names that were missing.

- **`bonus_xp on` / `off`** starts or ends a Bonus XP Weekend without a restart. `events.bonusExperience.enabled` was only read at player spawn, so changing it previously meant editing `game.properties`, restarting and having everyone log back in. The command changes the setting and brings everyone already online with it. `BonusExperience` also syncs on a settings reload now, the way `EvilTree` already does for its own event, so editing the file and running `reload settings` switches it over too, rather than leaving players on a multiplier the setting says they shouldn't have.

- **Bonus XP progress is scoped to the event it was earned in.** `bonus_xp_time` and `bonus_xp_counter` are persisted, so progress from a previous Bonus XP Weekend counted towards the next one, a player carrying ten hours online started a new event on the 1.1x floor. Players now keep the identity of the event they're taking part in (the same approach evil trees use for `evil_tree_spawn_id`) and progress is cleared when it doesn't match, while logging out and back in during the same event still keeps its decay. Applies to the settings-file path too, not just the console command.


engine/src/main/kotlin/world/gregs/voidps/engine/client/command/ConsoleLine.kt
```diff
@@ -0,0 +1,362 @@
...
+                Key.None
+            }
+            KILL_WORD -> {
+                kill(word(), cursor)
+                Key.None
+            }
+            TAB -> {
+                complete()
+                Key.None
+            }
```
GregHib, line 93: This doesn't seem to work in intellij + windows

Image

GregHib (commented):
This seems to be quite problematic when running as a dev 

Image

engine/src/main/kotlin/world/gregs/voidps/engine/client/command/ConsoleLine.kt
HarleyGilpin: Fixed in 9b617f777.

That key path only runs when the console has taken control of the terminal. Intellij's console isn't one: without raw mode the process is handed whole lines rather than key presses, so the tab arrives as a character in the middle of a line and there was nothing to complete on. It indents instead.

So completion now also answers a line that ends in a tab, printing the completion for the word in front of it rather than running it, since the line you meant is still ambiguous:

```
play	     -> players
bonus	    -> Matches:
                bonus_xp
                bonusxp
kick har	 -> kick "harley gilpin"
zzz	      -> No completions for 'zzz'.
```

Carrying on typing after a tab means it was only an indent, so `play\ters` runs `players`.

While in there: `play` on its own printed the entire command list, because it's three edits from `players` and the suggestion threshold is two. Half a name is a likelier mistake than a misspelling and only gets further away by edit distance the more of it is missing, so prefix matches are offered first now.

HarleyGilpin:
Tab is fixed in 9b617f777 — details on the inline thread. Short version: intellij's console can't be taken over, so the process gets whole lines instead of key presses and the tab lands in the middle of one. Completion now also answers a line ending in a tab, and `play` suggests `players` instead of printing every command.

One thing I haven't fixed, and can't from in here: your typed characters being cut through by log lines, as in the screenshot. Holding the input line below the log needs the terminal in raw mode, which needs a real one — on windows that means native console calls we'd need a dependency for, and intellij's console isn't a terminal at all so even those wouldn't help there. Running the jar from a terminal gets the pinned prompt; intellij gets plain lines.

If you'd rather it didn't run at all in that case, say and I'll make `console.enabled` default to starting only when there's a terminal to take, with `true` forcing the plain mode for piped input. That would keep dev runs free of the console entirely rather than half of one.

HarleyGilpin:
Windows gets the same console as everything else now, in 0501b00ca, but I can't test it, so it needs your eyes before this merges.

The dependencies are already packaged in our jar, pulled in by ktor, so the Jar size remains unchanged.

engine/src/main/kotlin/world/gregs/voidps/engine/client/command/ConsoleLine.kt
HarleyGilpin: Fixed on windows.

